PR #4, opened 2026-07-22
feat: add Pi session provider

Rousseau512:
## Summary

- add a registry-driven `pi` provider for standard Pi coding-agent v3 session JSONL
- project Pi trees into canonical messages, thinking, tool calls/results, summaries, metadata, usage, and sidechains
- resolve Pi session roots from `PI_CODING_AGENT_SESSION_DIR`, `PI_CODING_AGENT_DIR`, global `settings.json`, or the default `~/.pi/agent/sessions`
- add app and passive-pull CLI integration coverage
- update provider catalogs, package exports, skill references, and user documentation

## Implementation notes

- Pi IDs are provider-namespaced to avoid cross-source collisions.
- Modified sessions use tree-aware full replay with transactional `delete-session` cleanup and `countMode: 'total'`.
- Unchanged discovery checks file metadata before reading transcript contents.
- Image payloads are omitted from indexed/projected text; raw lookup remains bounded by the existing query API.
- Synthetic fixtures cover branching, parent causality, model state, mixed thinking/text/tool content, tool errors and file paths, usage, compaction and branch summaries, custom metadata, title fallback/clear semantics, bash status, torn writes, corrupt ancestry, replay cleanup, and raw lookup.

## Scope

This first version intentionally indexes only standard top-level files shaped `<project-dir>/<session>.jsonl`. Deeper nested Pi subagent run `session.jsonl` transcripts are excluded and documented for a follow-up.

Deleted-source reconciliation and recovery of provider roots created after app startup retain the current cross-provider behavior and are not patched with Pi-specific workarounds here.

## Validation

- `npm run typecheck`
- `npm run lint` — 0 errors; 4 pre-existing warnings
- `npm test` — 287/287 passing
- `npm run build:core`
- `npm run build:cli`
- focused Pi parser, app indexer, registry, session-detail, raw-query, and packaged CLI tests
- `git diff --check`

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -1,7 +1,7 @@
 # Obelisk
 
 Obelisk is explicit memory infrastructure for coding agents: it indexes local
-Claude Code, Codex, and Kimi Code sessions into a queryable SQLite evidence layer, and a
+Claude Code, Codex, Kimi Code, and Pi sessions into a queryable SQLite evidence layer, and a
 CodeAct runtime lets an agent write a small query, run it, and answer from real
 session history. This glossary pins the terms that are specific to Obelisk; it is
 not a spec.
@@ -29,7 +29,7 @@ promoted to an external tool surface.
 ## Indexing
 
 **Provider adapter**:
-A pure per-source module (claude, codex, kimi, later pi, …) that owns its
+A pure per-source module (claude, codex, kimi, pi, …) that owns its
 descriptor, watch roots, discovery, parsing, cursor interpretation, and raw
 record lookup. It discovers `IndexUnit`s rather than assuming one transcript
 file per unit; Kimi uses a session directory containing multiple wire logs. It
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 [![version](https://img.shields.io/github/v/tag/tommy0103/obelisk?label=version&style=flat-square)](https://github.com/tommy0103/obelisk/releases)
 [![license](https://img.shields.io/badge/license-AGPL--3.0-blue.svg?style=flat-square)](LICENSE)
 
-Past Claude Code, Codex, and Kimi Code sessions -- queryable by your agent, browsable by you.
+Past Claude Code, Codex, Kimi Code, and Pi sessions -- queryable by your agent, browsable by you.
 
 </div>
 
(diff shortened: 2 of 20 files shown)


tommy0103:
We’ll review the code before merging. (I might rebase some of the commits in this PR.) Thanks for your contribution!

As noted in issue #2, this is already referenced there. Glad to see this PR coming through!

tommy0103:
Hi @Rousseau512 — apologies for the long silence on this. The PR is still
on my list; I've been heads-down elsewhere and that's on me. Let me come
back with a proper review this week.

Separately, a heads-up you may want to look into on your own: a downstream
project (https://github.com/wutongyuonce/Trajex) shipped Pi session support
on 07-30. Its `packages/core/src/providers/pi.ts` is a different, much
shorter implementation overall, but a few structural pieces line up with
this PR closely enough that I thought you should know it exists — this is
your contribution, so it's your call entirely whether that matters to you.

Happy to share the specific diff if you want it. Either way it doesn't
affect the review here.

tommy0103:
Hi @Rousseau512 — thanks for this, and sorry it took a while to land on a decision.

We are going with #23 for Pi support, so I would like to close this one.

I want to be clear about why, because this was careful work and the decision is not a verdict on its quality. Both PRs implement the provider cleanly against the existing registry seam, both are well documented, and both come with a real test suite. What decided it was how much of Pi's on-disk format each one covers:

- **Session directory layouts.** Pi nests one directory per project under the default root, but a session directory configured through `--session-dir`, `PI_CODING_AGENT_SESSION_DIR`, or `sessionDir` in `settings.json` receives session files *directly*. This PR resolves those settings, but discovery only walks `<root>/<project>/*.jsonl`, so anyone with a configured session directory would index zero sessions.
- **Format versions.** This PR gates on `version === 3`. Pi migrates older sessions to v3 in memory and only rewrites the file when the session is actually resumed, so v1/v2 sessions a user never reopened stay invisible — and a future v4 would drop every Pi session at once. #23 covers v1 through v3.
- **Session identity.** Pi's `--session-id` lookup is project-local, so the header id alone is not globally unique. #23 keys on `(normalized header cwd, header id)`, which also keeps a session's identity stable across moves and copies.
- **Settings precedence.** Pi's project-level `.pi/settings.json` overrides the global one; this PR reads only the global file.
- **Tree semantics.** #23 handles `retainedTail` checkpoint compactions, orphan roots, and a durable `leaf`, and marks abandoned branches through `visibility` rather than overloading `is_sidechain`.

For what it is worth, the review notes from this PR became the checklist we ran #23 against, so the work here directly shaped what we are merging.

Thanks again for the effort — I hope you will send more our way.

PR #20, opened 2026-07-30
feat(app): open local file references at the referenced line

### LINKED ISSUE #18: [Sessions] Make local file references in Markdown open at the referenced line
## TL;DR

Agent transcripts are full of source references such as `app/src/main/index.ts:397`. Today they are either inert text or, when absolute, links that navigate the Electron window away from Obelisk entirely. Resolving them against the originating message's `cwd` and opening them at the referenced line would turn every path in a transcript into a working jump target.

## Problem description

Reading a session usually means moving between what the agent said and the code it was talking about. The references are already there and already precise — Claude Code and Codex both emit `path:line[:col]` routinely — but the reader has to copy the path out and find it by hand.

There is also a real defect in the same area: a Markdown link with a local href currently navigates the window, discarding all renderer state.

Prior exploration happened in #8, which is not being merged. The navigation defect was surfaced there.

## Current behavior

**Local links navigate away from the app.** `app/src/main/index.ts` installs neither a `will-navigate` handler nor a `setWindowOpenHandler`. A Markdown link with a `file:` href replaces the running SPA; recovering requires restarting the app.

**Relative references are never resolved.** `renderMarkdown()` (`app/src/renderer/src/utils.js:95-107`) passes hrefs through verbatim after `sanitizeMarkdown()`. A reference like `app/src/main/index.ts:397` therefore resolves against the renderer document URL, not against anything meaningful. This is the most common form in practice, so the majority of references in a transcript are unusable.

**The data needed to resolve them is already present but unread.** `messages.cwd` is stored per message and is already selected in the session query (`app/src/main/index.ts:397` and `:526`). No renderer code currently reads it.

**Line and column suffixes carry no meaning.** `:line[:col]` is part of the reference text today and is discarded by any consumer.

## Expected behavior

- Clicking a local file reference opens it in the user's editor at the referenced line and column.
- Relative references resolve against the `cwd` of the message they appear in.
- Clicking a local reference never navigates the Obelisk window away from the application.
- References that cannot be resolved stay readable and do nothing, rather than producing a broken navigation or an error dialog.

## Decision points

- [ ] **Editor scheme.** `vscode://file/{path}:{line}:{col}`, `cursor://`, or a configurable setting. A setting is more flexible but adds surface to Settings; a single default is simpler and can be widened later.
- [ ] **Whether bare relative paths in prose become links at all**, or only paths the Markdown source already wrote as links. Linkifying bare paths is far more useful and far more prone to false positives.

## Scope boundary

Hover preview of file contents is **out of scope**. It was explored in #8. Reading current disk contents sits badly with Obelisk being a historical browser: the reader is looking at a session from weeks ago, while the preview would show the file as it is today, with nothing marking the difference — and the contents the agent actually saw are already in the transcript as rendered tool results.

If a preview turns out to be worth having later, it should be scoped separately and would need to seek to the referenced line rather than read from the start of the file.

## Constraints

Transcript content is untrusted input — sessions are indexed from third-party logs. Two consequences worth keeping in mind:

- Resolved paths should stay within the session's `cwd` / `project_path` scope rather than allowing any absolute path on disk.
- `shell.openPath` runs `.app`, `.command` and `.sh` rather than opening them for viewing. If any fallback hands a path to the OS this way, its confirmation dialog must default to Cancel.

## Acceptance criteria

- [ ] Clicking a local file reference never navigates the Obelisk window away from the app.
- [ ] Relative references resolve against the originating message's `cwd`, not a session-level value — `cwd` can change within a single session.
- [ ] The `:line[:col]` suffix survives resolution and reaches the editor.
- [ ] References that cannot be resolved remain readable and trigger no navigation.
- [ ] Path-like text inside fenced code blocks is not turned into navigation accidentally.
- [ ] Resolved paths stay within the session's project scope; untrusted transcript content cannot cause an arbitrary path to be handed to the OS.
- [ ] Automated coverage for reference parsing and `cwd` resolution, following the existing root `tests/` + `node --test` pattern.
- [ ] `npm test` and `npm run typecheck` pass.

## Related

- #8 — prior exploration of this area, not merged. The `will-navigate` gap was surfaced there.
- #10 — wants the same extension point. Both features post-process `renderMarkdown()` output to make inline references interactive: #10 for `codex:<uuid>` session references resolving to internal navigation, this one for file paths resolving to an editor. Whoever gets there first should consider whether one mechanism can serve both, rather than two independent passes over the same HTML.

tommy0103:
Closes #18. Builds on #19 (merged) — the harness added here only runs under Electron 43.

## Why

Transcripts constantly point at source locations, and neither form works today.

Measured over the local index (104,071 assistant messages, 408 sessions):

| Form | Shape | Messages | Today |
|---|---|---|---|
| A | `[roadmap.md](/Users/me/proj/roadmap.md:162)` | 2,515 (119/408 sessions) | is an `<a>`; **navigates the window away**; line ignored |
| B | `` `src/hooks/hook.ts:40` `` in inline code | ~1,425 | not a link at all; path is relative |

Half of form A carries a `:line` suffix (1,253 messages) and nothing consumes it. Relative
*links* are effectively nonexistent — 4 messages in 104k — so form A needs no cwd and form B
needs it for everything.

## What

Both forms become anchors carrying the split-out components; the main process resolves them and
opens the configured editor at the line.

- Renderer owns detection and splits `path` / `line` / `column` / `endLine`; main never re-parses
  a raw string, so the pattern lives in one place.
- Form B requires **both** an extension and a line number, so `package.json`, `useState` and
  `npm run build` are never linkified. Text inside fenced `<pre>` blocks is skipped entirely.
- Only `file:` hrefs are percent-decoded. Decoding every href would corrupt a filename that
  legitimately contains `%`.
- Line ranges (`taskTree.ts:182-185`) parse alongside `:line:col`.
- Editor scheme is configurable in Settings (vscode / vscode-insiders / cursor / windsurf / zed),
  default vscode.

### Security

Transcript text is untrusted input — sessions are indexed from third-party logs. `file-ref:open`
derives the allowed roots from the database (`messages.cwd` and `sessions.project_path` **for
that session only**) and a renderer-supplied cwd is honoured only if it appears in that set.
Containment is re-checked after `realpath`, so a symlink cannot widen the reachable set.
Resolution is per session, never corpus-wide.

## Three things worth calling out in review

1. **Touches `packages/core`.** `SessionDetailMessage` gains `session_id` and `cwd`. The canonical
   assembly projected messages down to seven fields, so both were dropped before reaching the
   renderer — the feature did not work at all until this was fixed, and the unit suite stayed
   green throughout. Purely additive, but the type is shared with the CLI.

2. **Modifies an existing test.** The Electron double in `tests/app-main-settings.test.mjs` lacked
   `setWindowOpenHandler` and `getURL`, which the new `createWindow` calls. All 11 mocks were
   filled in rather than guarding the production code with `typeof === 'function'` checks.

3. **Changes behaviour for external links too.** Installing a proper navigation guard means http(s)
   links in a transcript now open in the system browser instead of replacing the SPA. That is
   outside the file-reference feature proper, but it is the same one-line defect.

## Known gap

Memory bodies have no cwd or session, so references there stay unmarked rather than looking
actionable and doing nothing. Wiring Memory up is worth a separate issue.

## Verification

- `npm test` — 293/293
- `npm run typecheck`
- `npm run lint` — unchanged from the `main` baseline
- `test:electron` 15 · `test:electron:timeline` 55 · `test:electron:reader-state` 11 ·
  `test:electron:file-refs` 17 — 98 assertions

The new harness drives the real renderer and asserts marking, the `<pre>` exclusion, the data
attributes, the click payload, and that clicking never navigates the window.

### DIFF
--- a/app/electron.vite.config.ts
+++ b/app/electron.vite.config.ts
@@ -13,6 +13,7 @@ export default defineConfig({
       rollupOptions: {
         input: {
           index: resolve('src/main/index.ts'),
+          'file-reference': resolve('src/main/file-reference.ts'),
           indexer: resolve('src/main/indexer.ts'),
           'indexer-service': resolve('src/main/indexer-service.ts'),
           'indexer-worker': resolve('src/main/indexer-worker.ts'),
--- a/app/package.json
+++ b/app/package.json
@@ -21,7 +21,8 @@
     "verify:icons": "node scripts/verify-icons.mjs",
     "test:electron": "electron-vite build && electron --no-sandbox tests/electron-concurrency.mjs",
     "test:electron:timeline": "electron-vite build && electron --no-sandbox tests/electron-session-virtualization.mjs",
-    "test:electron:reader-state": "electron-vite build && electron --no-sandbox tests/electron-session-reader-state.mjs"
+    "test:electron:reader-state": "electron-vite build && electron --no-sandbox tests/electron-session-reader-state.mjs",
+    "test:electron:file-refs": "electron-vite build && electron --no-sandbox tests/electron-file-references.mjs"
   },
   "build": {
     "appId": "com.obelisk.app",
--- a/app/src/main/file-reference.ts
+++ b/app/src/main/file-reference.ts
@@ -1,0 +1,90 @@
+import fs from 'node:fs';
+import path from 'node:path';
+
+const EDITOR_SCHEMES: Record<string, string> = {
+  vscode: 'vscode',
+  'vscode-insiders': 'vscode-insiders',
+  cursor: 'cursor',
+  windsurf: 'windsurf',
+  zed: 'zed',
+};
+
+const DEFAULT_EDITOR_SCHEME = 'vscode';
+
+interface FileReferenceQuery {
+  rawPath?: string | null;
+  cwd?: string | null;
+  roots?: (string | null | undefined)[];
+}
+
+interface EditorUrlOptions {
+  scheme?: string | null;
+  filePath: string;
+  line?: number | null;
+  column?: number | null;
+}
+
+function normalizeRoots(roots: (string | null | undefined)[] = []): string[] {
+  const seen = new Set<string>();
+  for (const root of roots) {
+    if (typeof root !== 'string' || !root.trim() || !path.isAbsolute(root)) continue;
+    seen.add(path.resolve(root));
+  }
+  return [...seen];
+}
+
+function isWithinRoots(candidate: string, roots: string[]): boolean {
+  return roots.some((root) => candidate === root || candidate.startsWith(root + path.sep));
+}
+
+// A reference may be absolute, relative to the message cwd, or project-relative with a
+// leading slash (`/src/foo.ts`). path.join treats the leading slash as a no-op, so the same
+// join covers the last two cases.
+function fileReferenceCandidates({ rawPath, cwd, roots = [] }: FileReferenceQuery): string[] {
+  if (typeof rawPath !== 'string' || !rawPath.trim()) return [];
+  const cleaned = rawPath.trim();
+  const bases = normalizeRoots([cwd, ...roots]);
+  const candidates: string[] = [];
+  if (path.isAbsolute(cleaned)) candidates.push(path.resolve(cleaned));
+  for (const base of bases) candidates.push(path.resolve(base, `.${path.sep}${cleaned}`));
+  return [...new Set(candidates)];
+}
+
+// Resolves to an existing file that lives inside one of the session's own roots. Transcript
+// text is untrusted input, so containment is checked after realpath — a symlink pointing out
+// of the project must not widen what the app will open.
+function resolveFileReference(query: FileReferenceQuery): string | null {
+  const roots = normalizeRoots([query.cwd, ...(query.roots || [])]);
+  if (!roots.length) return null;
+  for (const candidate of fileReferenceCandidates(query)) {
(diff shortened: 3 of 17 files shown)


tommy0103:
Two defects found while exercising the Memory view, both fixed in the latest push.

**Same-origin URLs were handed to the system browser.** A relative link in a Memory body
(`../../docs/trajectory-v0-design.md`) resolves against the renderer origin, so the navigation
guard correctly prevented it but then treated it as external — in dev that opened the system
browser on a Vite-served source file. `releaseNavigation` now only hands off genuinely
cross-origin targets. External links still open in the browser (verified separately).

**The marking pass measurably destabilised the timeline harness.** `markFileReferences` ran two
`querySelectorAll` calls for every rendered row, which was enough to lose a cold-open layout race
in `test:electron:timeline`:

| branch | timeline harness |
|---|---|
| `main` | 7 pass / 7 runs |
| this branch, before | 5 pass / 8 runs |
| this branch, after | 5 pass / 5 runs |

`renderMarkdown` already holds the rendered HTML as a string, so a `String.includes` pre-filter
now decides whether the DOM walk is worth doing at all. Most messages contain neither a link nor
inline code and skip it entirely.

Re-verified: `npm test` 295/295, typecheck, and all four harnesses (15 / 55 / 11 / 17).

Still out of scope, and worth its own issue: Memory bodies link to docs relative to the memory
file's own directory (`memories.path`), which is a different resolution base than a session cwd.
Those references remain unmarked.

PR #23, opened 2026-08-02
feat(core): add first-class Pi session indexing

SaladDay:
## Summary

Depends on #35.

[Pi-only diff (`96d8523...349508a`)](https://github.com/SaladDay/obelisk/compare/feat/provider-indexing-foundations...feat/pi-session-support)

This PR adds Pi as a first-class Obelisk provider while keeping Pi-specific semantics inside one adapter.

- Resolve the automatic session root with Pi 0.83 precedence: `PI_CODING_AGENT_SESSION_DIR`, merged global/project `sessionDir`, then `~/.pi/agent/sessions`.
- Discover and parse Pi v1-v3 sessions, including messages, images, tools, bash records, custom messages, errors, summaries, model usage, and exact raw JSONL evidence.
- Reconstruct Pi's selected context from `leaf`, orphan roots, branch summaries, legacy `firstKeptEntryId` compaction, and checkpoint `retainedTail`.
- Mark superseded branch evidence `inactive` and display-suppressed transport evidence `hidden`; the app stays active-only while agents may request inactive evidence explicitly.
- Use `(normalized header cwd, header id)` as logical identity and source paths as provenance, preserving identity across copies and moves while handling replacements, unlinks, and divergent duplicates safely.
- Bind raw lookup to the committed source cursor so replaced content cannot be exposed as stale evidence.
- Follow Pi's tolerance for malformed physical JSON lines, while JSON-valid structural errors fail the affected unit atomically.

Shared code only registers the provider and threads the launch `cwd` into automatic Pi root resolution; there are no `source === "pi"` branches in indexing, persistence, querying, settings, or rendering.

## Compatibility

| Component | Version |
|---|---|
| Obelisk app | `0.2.2` |
| Obelisk core | `0.2.0` |
| Obelisk CLI | `0.2.2` |
| Pi coding agent | `@earendil-works/pi-coding-agent@0.83.0` |
| Pi agent core | `@earendil-works/pi-agent-core@0.83.0` |
| Pi session format | v1, v2, v3 |
| Base stack commit | `96d8523` |
| Validation runtime | Node `24.1.0`, npm `11.3.0`, macOS |

The implementation was checked against Pi's versioned [settings documentation](https://github.com/earendil-works/pi/blob/v0.83.0/packages/coding-agent/docs/settings.md), [session format](https://github.com/earendil-works/pi/blob/v0.83.0/packages/coding-agent/docs/session-format.md), coding-agent context projection, and agent-core retained-context storage.

## Verification

- `npm test` — **408/408 passed**
- `npm run typecheck` — passed
- `npm run build:core` — passed
- `npm run build:cli` — passed
- `npm run build:skill` — passed
- `npm run lint` — 0 errors; 4 existing warnings
- `git diff --check` — passed

Installed Pi 0.83 runtime probes covered official `SessionManager` persistence and context projection, global/project settings precedence, malformed settings fallback, tool calls/results, image and error records, cache usage, compaction, branch selection, moves, duplicate copies, missing inventories, and structurally invalid sessions. Targeted regressions also confirm that a malformed Pi file retries alone and never forces unrelated unchanged sessions to reparse.

### Randomized differential check

`tests/pi-randomized-differential.test.mjs` runs unconditionally against vendored, version-pinned Pi 0.83 context oracles.

With seed `0x5eedc0de`, **512/512** generated sessions matched:

- 18,124 entries and 11,586 messages
- 1,457 leaf changes, including 167 null leaves
- 2,230 compactions, including 1,137 retained-tail compactions
- 2,191 branch summaries
- 1,060 orphan parent edges

The comparison checks ordered canonical evidence, including role/content, summary source/content, and retained-tail placement.


SaladDay:
Would this visibility model work?

- `visible`: returned by default
- `inactive`: excluded by default; available with `includeInactive: true`
- `hidden`: never returned by standard helpers

`includeInactive` would apply to all visibility-aware helpers. Pi would mark records outside its effective context as `inactive`; internal or display-suppressed records would remain `hidden`. No schema migration is needed; a Pi index-marker bump would replay existing sessions. The app would remain active-only by default.

SaladDay:
Follow-up scope:

This PR would cover Pi's three-state visibility and agent opt-in. Claude Code fork/session relationships and app branch browsing can remain separate follow-ups. Inactive Pi messages, summaries, and raw evidence would be queryable now; structured inactive tool pairs can follow where pairing is unambiguous.

tommy0103:
Both of the blocking items from the last round check out on `3784306`. Verified
locally: `npm test` 384/384, typecheck clean, lint 0 errors.

**Marker retry** — confirmed fixed, and the cursor deletion for failed units is a
better fix than simply relaxing the gate. My earlier reasoning was that a failed
unit never committed a cursor so it stays retryable on its own; that misses the
case where the file succeeded on an earlier run and still holds a stale cursor.
Deleting it explicitly is the right call.

**Differential oracle** — the transcription is faithful. Since a hand-transcribed
oracle can in principle encode the same misreading on both sides, I checked it
independently: 400 randomly generated trees (my own generator, different
distribution from yours — compactions with and without `retainedTail`, branch
summaries, orphan parent edges) run through both the vendored oracle and the
`buildContextEntries()` of a locally installed Pi, **400/400 matched**. My local
Pi is 0.82.0 rather than 0.83.0, so this corroborates rather than replaces your
own cross-check, but it is independent evidence that the pinned oracle is sound.

---

There is one regression I would like to flag before this lands.

## Fail-closed inventory takes the whole provider offline, permanently

`createProviderIndexPlan` now skips every unit when the inventory cannot be
enumerated:

```js
if (markerMissing && !inventoryComplete) continue;
```

Because `pendingMarkers` is only populated when `inventoryComplete`, the marker
is never written, so `markerMissing` stays true and the state never resolves.

Repro — a single unreadable directory anywhere under the provider root:

```js
const locked = join(root, 'locked');
mkdirSync(locked, { recursive: true });
chmodSync(locked, 0o000);          // EACCES on readdirSync -> complete: false
```

Measured against a fresh database, three consecutive builds:

```
run1: planned=0 committed=0 marker=0 incomplete=[pi]
run2: planned=0 committed=0 marker=0 incomplete=[pi]
run3: planned=0 committed=0 marker=0 incomplete=[pi]
sessions indexed: 0
```

Same probe on `main` for comparison, using the Claude provider and one
unreadable directory under `~/.claude/projects`:

| | result |
|---|---|
| `main` | `planned=1 sessions=1` — the unreadable directory is skipped, everything else indexes |
| this PR | `planned=0 sessions=0` — nothing is indexed, and it never recovers |

Since all four adapters now declare `snapshotIntegrity: 'strict'`, this is not
Pi-specific. The blast radius is bounded by whether the marker already exists:

```
fresh install (no marker):        planned=0 sessions=0
existing install (marker present): planned=1 sessions=1
```

The Claude, Codex and Kimi markers are unchanged in this PR, so existing
installations keep working for those three and only Pi is affected. A **fresh
install is the bad case**: a new user with any unreadable directory under
`~/.claude/projects` gets an empty database that never fills, on the primary
provider.

It is also silent. The `incomplete_snapshot` early return only fires when
`force` is true; on the ordinary background path the build plans zero units and
returns normally, and nothing in the renderer consumes `incompleteProviders`.
What the user sees is a successful build and an empty library.

Realistic triggers are not exotic: a directory owned by another user, a stale
mount, a dangling symlinked directory, or an Electron build without Full Disk
Access on macOS when the root sits under a protected path.

## Suggested shape

The reasoning behind fail-closed is right — a written marker is a claim that the
whole provider was replayed under the new semantics, and you cannot honestly make
that claim from a partial inventory. The issue is that two separate consequences
got fused into one:

- *cannot certify a complete inventory* → **do not write the marker** ✅
- *cannot certify a complete inventory* → **do not index anything** ← this is the
  part that turns a recoverable state into a permanent one

Decoupling them keeps the guarantee and removes the deadlock:

1. Index whatever units are readable, even when the inventory is incomplete.
2. Withhold the marker in that case, exactly as now.
3. Surface `incompleteProviders` on the ordinary path too, with the directory
   that could not be read.

The worst case then degrades to "the replay is attempted again next launch" —
wasteful, but correct, and self-healing once the directory becomes readable.

A regression test worth pinning: *with an unreadable directory under the root, the
readable sessions are still indexed.*

Happy to be talked out of this if I am misreading the intent of the
`requireCompleteSnapshot` contract — but as it stands a fresh install can end up
with a permanently empty index and no error.

SaladDay:
Agreed. `requireCompleteSnapshot` should gate snapshot certification and destructive publication, not suppress readable units during an ordinary build.

I'll decouple those paths: ordinary builds will index readable units, report the incomplete inventory, and withhold the provider marker. Force rebuilds will remain fail-closed. I'll also carry the unreadable path through a provider-neutral diagnostic and add a recovery regression test.

tommy0103:
The three-state model is the right call, and it is better than the `includeHidden`
flag I originally suggested. That suggestion conflated two things your split keeps
apart: abandoned-but-real content, and records that were never conversation in the
first place. With a single flag, an agent that wants to read an abandoned Pi branch
would also have to pull in Codex's `<environment_context>` envelopes. `inactive`
and `hidden` sharing a default but not a recovery path is the distinction that
makes the opt-in meaningful.

One gap in the proposal as written, and I think it is load-bearing.

## `visibility` has to be in the result, not only in the predicate

`query.ts:138` selects `uuid, session_id, text, content_type, is_meta, role,
timestamp, model, cwd, source` — no `visibility`. So `includeInactive: true`
returns a set the agent cannot partition. It will find "we decided to use approach
A", not know that path was abandoned, and report it as history.

That is worse than the current unreachability, which is at least honest. Two things
would close it: return the column, and say in the skill docs how an inactive hit
should be phrased ("you tried X and moved away from it", not "you used X").

## Which adapters can actually populate `inactive`

Before assuming the other adapters should follow Pi here, I checked what their
formats can express. Two layers matter: whether a tree exists at all, and whether
the format records the act of abandoning one.

At the code level, only two adapters read a parent pointer from their source:

```
claude:  parent_uuid: obj.parentUuid        <- from the source format
codex:   parent_uuid: sm.lastMessageUuid    <- synthesized ("previous message")
kimi:    parent_uuid: previousUuid          <- synthesized ("previous message")
pi:      parent_uuid: parent                <- computed from the source tree
```

Against real local transcripts:

| source | files / lines | top-level fields | parent-pointer fields |
|---|---|---|---|
| Claude | 25 / 23,236 | `type, sessionId, timestamp, parentUuid, isSidechain, uuid, …` | `parentUuid` × 17,043 |
| Codex | 25 / 22,148 | `timestamp, type, payload` | none |
| Kimi | 19 / 2,463 | `type, time, event, model, …` | none |

I probed for `parentUuid`, `parentId`, `parent_id`, `parent`, `leaf`, `leafId`,
`branch`, `fromId`. Codex and Kimi are append-only event logs; their parent chains
in Obelisk are entirely adapter-synthesized. So:

| | tree shape | leaf pointer | branch operation recorded | active path derivable |
|---|---|---|---|---|
| **Pi** | yes (`id`/`parentId`) | yes, durable | yes (`branch_summary.fromId`) | yes |
| **Claude Code** | yes (`parentUuid`) | no | no | no — inference only |
| **Codex** | no | — | — | — |
| **Kimi** | no | — | — | — |

Claude has the shape but not the semantics. I looked for an authoritative rewind
signal and did not find one: `promptSource`, `origin`, `entrypoint`,
`permissionMode` all describe how a prompt was entered (`typed`, `cli`, `sdk-ts`,
`{kind: human}`), not whether a branch was abandoned on purpose. They appear on
both sides of a fork.

## The criterion is not "has a tree"

Kimi has no tree, but it does record supersession explicitly: `context.undo`
retracts the last N real user prompts, all-or-nothing, refusing to cross a
compaction boundary. That is an authoritative "these were retracted" record — the
same class of fact as a Pi branch, in a stack rather than a tree.

That suggests the contract should be about attestation:

> `inactive` means the provider can confirm from its source that the record was
> explicitly superseded. The form is provider-specific: a branch in Pi, an undo in
> Kimi. Providers whose format does not record supersession leave everything
> `visible`.

Which lands as:

- **Pi** — branch operations. Populates `inactive`.
- **Kimi** — `context.undo` / `context.clear`. Could populate `inactive`; today the
  adapter deletes those records instead. (Related fidelity gap: #26.)
- **Claude Code** — nothing to attest with. Stays `visible`.
- **Codex** — no branching exists. N/A rather than pending.

## Why Claude should not mirror Pi

I suggested earlier that Claude's abandoned branches should be marked `inactive`.
Having looked at the data, I do not think that holds.

Across 43 local transcripts I found 208 branches off the surviving path. Roughly
56% were multi-message and 41% were single user messages with no reply at all —
someone typed, then rewrote before sending. Median abandoned-branch length was 2,
p90 was 18. Even inside the multi-message group the heads were not all user
decisions; some were harness-injected (`<task-notification>` starting a 5-message
branch). Treat these percentages as a rough cut — my classifier counted nested
forks inside already-dead regions separately, so the absolute numbers run high. The
shape is solid; the precision is not.

Two consequences. Marking these `inactive` would require a heuristic (a length
threshold, filtering injected heads), and `types.ts` says of this exact field:
"Provider-normalized display eligibility. Assemblers never infer this from text."
And it would be a retrieval regression: those branches are searchable today, so
reclassifying them wholesale moves real content behind an opt-in on the strength of
a guess.

The genuine Claude problem is a different one and needs no new semantics: abandoned
and active branches interleave by timestamp in the assembled timeline, and
`session-detail.ts` merges consecutive assistant `tool_use` messages across them.
`parent_uuid` is already indexed; that is a presentation fix.

## Scope

This supports the split in your follow-up comment, for a firmer reason than scope:
Claude cannot currently express what Pi expresses. A follow-up issue is better
framed as "blocked until Claude Code's transcript records rewinds" than "not yet
implemented."

That makes the capability matrix the one piece I would want to land **with** this
PR rather than after it. Not as "`inactive` is Pi-only for now, others to follow",
but as the format-capability statement above. The difference matters: with the
first wording an agent reads an empty `inactive` result on a Claude session as
"nothing was abandoned here" — which the 43-transcript sample says is false about
39% of the time. With the second, it knows that is the most that source can say.

SaladDay:
Agreed. I’ll return `visibility` from every `includeInactive`-aware helper and document `inactive` as source-attested supersession, including the provider capability matrix. Pi will populate it; Claude stays visible, Codex is N/A, and Kimi remains a follow-up behind #26.

SaladDay:
Implemented the agreed follow-up: source-attested visible/inactive/hidden retrieval, visible-only session and app surfaces and counts, safe publication under partial inventories, and exact provider/path/error diagnostics in the CLI and Settings. Pi loading was rechecked against 0.83.0, including malformed-line behavior. Validation: 396/396 tests; typecheck and core/CLI/skill builds pass; lint has 0 errors and 4 unchanged warnings; diff check is clean.

SaladDay:
Merged current `main` (`1941e64`) and resolved the provider, indexer, and Settings integration conflicts. The branch now also shares configured provider roots across indexing, query, and raw lookup; keeps inactive Pi tool evidence behind `includeInactive`; withholds cross-provenance replacement during an incomplete identity census; and scopes Pi raw evidence to one source-message container.\n\nFinal verification: 414/414 root tests passed; typecheck and the core, CLI, and skill builds passed; lint has 0 errors and 4 unchanged warnings. Electron is 4/5 because the upstream session-virtualization suite remains timing-sensitive; a clean main control also hung in that suite.

SaladDay:
I split the change into two linear commits and PRs: #35 contains the provider-neutral indexing foundations, and this PR contains the Pi adapter, registration, tests, and docs. The Pi-only diff is https://github.com/SaladDay/obelisk/compare/feat/provider-indexing-foundations...feat/pi-session-support.

Issue #5, opened 2026-07-22
Feature request: Support pluggable custom provider packages


yuchanns:
Thanks for creating and sharing Obelisk—it is an excellent memory skill.

I would like to request an extension mechanism for custom providers.

## Motivation

Obelisk can already be extended with another provider, but doing so currently requires modifying or forking the skill itself and registering the provider in its source tree.

I am doing this for my custom agent, where we maintain a customized Obelisk fork with a provider for my own conversation store. The integration works, but it couples the provider to Obelisk's internals:

Every upstream update requires us to reapply or rebase the provider changes.
The provider cannot be versioned and distributed independently.
Application-specific dependencies and configuration become part of the Obelisk fork.

## Proposed capability

Allow users to install trusted, local provider packages and register them with Obelisk through a supported extension point, without modifying the Obelisk package itself.

I do not have a strong preference for the exact discovery mechanism. It could, for example, use an explicit provider list in configuration, a manifest, or a dedicated local providers directory. The important properties would be:

- Providers are loaded only when explicitly configured; Obelisk does not automatically download or execute untrusted packages.
- The provider interface and lifecycle are small, stable, and documented.
- Provider identity and compatibility are validated with clear error messages.
- Provider-specific dependencies and configuration remain outside Obelisk core.
- Custom providers participate in the same indexing and retrieval flows as built-in providers.
- A broken custom provider fails clearly without corrupting the existing index or silently disabling other providers.
- A minimal example provider package would also make the extension contract easier to adopt.

tommy0103:
Thanks for your supporting!
We really appreciate this constructive issue. Point taken – we'll give some thought to how we can build this.

PR #29, opened 2026-08-03
fix: preserve utf8 across JSONL buffer boundaries 保留跨 JSONL 缓冲区边界的 UTF-8 字符

wutongyuonce:
## What and why

Fix shared JSONL line reading when a multi-byte UTF-8 character spans the 64 KiB read buffer boundary.
readLines() previously decoded each buffer independently with Buffer#toString('utf8'). A split character could become replacement characters (�), causing equivalent Codex event_msg and response_item text to differ and bypass assistant-message deduplication.
修复多字节 UTF-8 字符跨越 64 KiB 读取缓冲区边界时，共享 JSONL 行读取器错误解码的问题。
此前 readLines() 对每个 buffer 独立调用 Buffer#toString('utf8')。当字符被拆到两个 buffer 中时，会被替换为 �；因此等价的 Codex event_msg 与 response_item 文本不同，导致 assistant 消息去重失效。
1. Use Node's StringDecoder('utf8') in the shared readLines() helper. 在共享 readLines() 中使用 Node 内置 StringDecoder('utf8')。
2. Flush pending bytes with decoder.end() after the final read. 在读取完成后调用 decoder.end()，刷新残留字节。
3. Preserve the existing newline splitting and remainder handling. 保持原有的按换行拆分与 remainder 拼接逻辑。
4. Add a regression test that places 中 across the 64 KiB boundary and verifies Codex emits one deduplicated assistant message. 新增回归测试：将 中 精确放在 64 KiB 边界，验证 Codex 最终只产出一条去重后的 assistant 消息。
All consumers of the shared JSONL reader now preserve UTF-8 correctly across buffer boundaries, including Claude and Codex parsing paths. No Codex-specific deduplication logic changed.
所有使用共享 JSONL 读取器的路径都会正确保留跨 buffer 的 UTF-8 字符，包括 Claude 与 Codex 的解析流程。未修改 Codex 专用的去重逻辑。

## Verification

- [x] `npm test` — 309 pass / 0 fail
- [x] `npm run lint` — 0 errors
- [x] `npm run test:electron:all` — 5/5 suites (app/ unchanged in this PR; reviewer-verified. Locally 4/5: electron-session-virtualization's final renderer-timing wait times out only under this sandboxed terminal)
- [ ] New tests actually run in CI (`.github/workflows/`) 
- [x] Every capability described above was exercised end to end from the
      outermost entry point (including anything shown in a screenshot)

## Migration note

`CLAUDE_CANONICAL_TRANSCRIPT_MARKER` (`__claude_canonical_transcript_v3__`) and
`CODEX_CANONICAL_TRANSCRIPT_MARKER` (`__codex_canonical_transcript_v3__`) were bumped so that
already-indexed sessions are re-read. Without the bump, discovery's mtime short-circuit would leave
rows with the corrupted � text and duplicated assistant messages in place forever. Cost: every
existing installation performs one full re-index on next discovery. Kimi is unaffected — it reads
whole files, not via `readLines`.

## Deliberately out of scope

Deliberately left out: provider-specific parser changes. The defect is in the shared reader, so fixing it once covers every JSONL provider.
刻意未改动各 provider 的解析逻辑。根因位于共享读取器，在此处统一修复即可覆盖所有 JSONL provider。

### DIFF
--- a/packages/core/src/parsing.ts
+++ b/packages/core/src/parsing.ts
@@ -1,10 +1,11 @@
 // Core's pure parse/discover helpers — node:sqlite-free by construction, so the compiled
 // providers can be consumed by the app (better-sqlite3 / a Node without
 // node:sqlite). Originally extracted verbatim from db/indexer; it now exposes a
-// typed seam while remaining limited to node:fs/path/os.
+// typed seam while remaining limited to node:fs/os/path/string_decoder.
 import { closeSync, existsSync, openSync, readSync, readdirSync, statSync } from 'node:fs';
 import { homedir } from 'node:os';
 import { isAbsolute, join, normalize } from 'node:path';
+import { StringDecoder } from 'node:string_decoder';
 
 const CLAUDE_DIR = join(homedir(), '.claude');
 const CODEX_DIR = join(homedir(), '.codex');
@@ -102,22 +103,25 @@ function filePath(name: string, input: JsonRecord | null | undefined): string | 
 
 function isDir(p: string): boolean { try { return statSync(p).isDirectory(); } catch { return false; } }
 
+const READ_BUFFER_SIZE = 64 * 1024;
+
 function readLines(filePath: string, callback: (line: string) => boolean | void): void {
   const fd = openSync(filePath, 'r');
-  const bufSize = 64 * 1024;
-  const buf = Buffer.alloc(bufSize);
+  const buf = Buffer.alloc(READ_BUFFER_SIZE);
+  const decoder = new StringDecoder('utf8');
   let remainder = '';
   let bytesRead;
   try {
-    while ((bytesRead = readSync(fd, buf, 0, bufSize, null)) > 0) {
-      const chunk = remainder + buf.toString('utf8', 0, bytesRead);
+    while ((bytesRead = readSync(fd, buf, 0, READ_BUFFER_SIZE, null)) > 0) {
+      const chunk = remainder + decoder.write(buf.subarray(0, bytesRead));
       const lines = chunk.split('\n');
       remainder = lines.pop() ?? '';
       for (const line of lines) {
         if (line && callback(line) === false) return;
       }
     }
-    if (remainder) callback(remainder);
+    const tail = remainder + decoder.end();
+    if (tail) callback(tail);
   } finally {
     closeSync(fd);
   }
@@ -366,7 +370,7 @@ function codexToolOutput(payload: JsonRecord): string | null {
 }
 
 export {
-  CLAUDE_DIR, CODEX_DIR, PROJECTS_DIR, CODEX_SESSIONS_DIR, TEXT_LIMIT,
+  CLAUDE_DIR, CODEX_DIR, PROJECTS_DIR, CODEX_SESSIONS_DIR, TEXT_LIMIT, READ_BUFFER_SIZE,
   trunc, truncJson, extractText, extractContentType, extractMessageIsMeta, isSkillInstructions, filePath, isDir, readLines,
   legacyProjectPathFromSlug, normalizeObservedCwd, projectSlugFromPath, inferProjectPath,
   discoverJsonlFiles, discoverCodexJsonlFiles,
--- a/packages/core/src/providers/claude.ts
+++ b/packages/core/src/providers/claude.ts
@@ -35,7 +35,7 @@ function cursorToSkip(cursor: Cursor): number {
 }
 
 export const name = 'claude';
-export const CLAUDE_CANONICAL_TRANSCRIPT_MARKER = '__claude_canonical_transcript_v2__';
+export const CLAUDE_CANONICAL_TRANSCRIPT_MARKER = '__claude_canonical_transcript_v3__';
 
 interface ClaudeWorkflowUnitMeta {
   readonly kind: 'workflow';
--- a/packages/core/src/providers/codex.ts
+++ b/packages/core/src/providers/codex.ts
@@ -35,7 +35,7 @@ import type {
 } from './types.ts';
 
 export const name = 'codex';
-const CODEX_CANONICAL_TRANSCRIPT_MARKER = '__codex_canonical_transcript_v2__';
+const CODEX_CANONICAL_TRANSCRIPT_MARKER = '__codex_canonical_transcript_v3__';
 
 const HIDDEN_CONTEXT_ENVELOPE_RE = /^\s*<(environment_context|codex_internal_context)\b[^>]*>[\s\S]*<\/\1>\s*$/;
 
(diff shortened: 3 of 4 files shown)


tommy0103 (changes requested):
Thanks — the diagnosis is right, the fix is in the right place, and the test has teeth. I verified all three before writing this.

**The defect reproduces exactly as described.** Running your test against an unfixed `parsing.ts`:

```
actual:   [ '���', '中' ]
expected: [ '中' ]
```

The 3-byte `中` straddles the boundary, each buffer decodes independently, and the `event_msg` text stops matching the `response_item` text, so dedup lets both through. With the fix: 309 pass / 0 fail, typecheck clean, `build:cli` / `build:core` / `electron-vite build` all fine, `test:electron:all` 5/5.

**Scoping the fix to the shared reader is correct.** I grepped `packages/core/src`, `packages/cli/src` and `app/src`: `parsing.ts` held the only manual buffer decode, so there is genuinely nothing provider-specific to change. `kimi.ts` reads whole-file via `readFileSync(path, 'utf8')` and was never affected; `claude.ts` and `codex.ts` both go through `readLines` and both are fixed by this.

---

## Blocking: `indexVersionMarker` is not bumped

`CONTRIBUTING.md` names this one directly:

> **Bump `indexVersionMarker` whenever you change uuid format, role normalization, or anything else affecting already-stored rows.** Otherwise the mtime short-circuit in discovery leaves old-format rows in the database forever.

This PR changes what `parse()` produces for sessions that are already indexed. Per `packages/core/src/provider-indexing.ts:38-46`, a full re-index only happens when the provider's marker row is absent from `index_state`; otherwise discovery takes the stored cursor and the file is never re-read.

So as it stands, the fix only applies to sessions indexed *after* the upgrade. Every session already in the database keeps both symptoms this PR exists to remove — the `���` text **and** the duplicated assistant messages.

Both affected providers need their marker bumped in the same commit:

- `CLAUDE_CANONICAL_TRANSCRIPT_MARKER` — `packages/core/src/providers/claude.ts:40`, currently `__claude_canonical_transcript_v2__`
- `CODEX_CANONICAL_TRANSCRIPT_MARKER` — `packages/core/src/providers/codex.ts:40`, currently `__codex_canonical_transcript_v2__`

(Kimi's marker should stay where it is — it does not use `readLines`.)

This does force one full re-index for every existing installation. That cost is real and worth calling out in the PR description, but the alternative is shipping a fix that leaves the corruption in place for everyone who already has data.

## Needs a rebase

`main` moved: `1941e64` replaced the `createRequire` pattern with named imports, and it rewrote the same lines you are changing. `git merge-tree` reports a content conflict in `packages/core/src/parsing.ts`, and GitHub now shows this PR as `CONFLICTING`.

The bug itself survived that refactor — `buf.toString('utf8', 0, bytesRead)` is still there — so this PR is still needed. After rebasing the loop reads:

```ts
while ((bytesRead = readSync(fd, buf, 0, bufSize, null)) > 0) {
```

Note that per the verification contract, everything you re-check has to be re-run on the rebased head; the results above are mine, on the pre-rebase commit.

## Verification section

Two boxes are ticked with the template placeholders still in them:

```
- [x] `npm test` — <!-- N pass / 0 fail -->
- [x] `npm run test:electron:all` — <!-- required if app/ changed; N/N suites -->
```

The contract asks for the numbers, not the checkmark. (For what it is worth the numbers are real — I got 309/0 and 5/5 — but filling them in is the author's half of the contract.)

One box is not accurate through no fault of yours: **`tests/codex-parse.test.mjs` does not run in CI.** `.github/workflows/cli.yml` runs a four-filename whitelist and this is not one of them, so this regression can never go red on a pull request. That is a pre-existing repository gap, now tracked in #34 — please just untick that box rather than trying to solve it here.

## Fixture should be real provider output

Verification contract #5: *Fixtures are real provider output, not hand-written approximations.*

The fixture injects a field that no real Codex `agent_message` payload carries:

```js
payload: { type: 'agent_message', padding: '', message: '中' }
```

Padding the `message` itself (and the matching `output_text` in the `response_item`) reaches the same byte offset, keeps the fixture shaped like real output, and has the side benefit of exercising dedup on a long assistant message rather than a one-character one.

## The boundary constant is decoupled from the code under test

The test hardcodes `65535`; `parsing.ts` declares `const bufSize = 64 * 1024`. Nothing connects them. If the buffer size is ever changed, `中` is no longer on a boundary and **the test keeps passing without testing anything** — it would pass against the unfixed reader too.

Deriving the offset from the reader's buffer size, or asserting that `中`'s byte offset really does land on the boundary, keeps the fixture from silently degenerating. (Same spirit as the "no assertions with sub-pixel headroom / self-calibrate" rule in the renderer section.)

## Nit

`packages/core/src/parsing.ts:4` still says the module is *"limited to `node:fs/path/os`"*. Worth updating in the same commit now that `node:string_decoder` is in there.

For the record: I initially flagged the static `import` as inconsistent with the file's `createRequire` style — that was wrong, and it is moot after `1941e64` anyway. Static named imports are now the convention here.

---

To summarise: bump the two markers, rebase, re-run and paste the numbers, and tighten the fixture. The core change itself I would take as-is.

wutongyuonce:
Thanks for the thorough review — all points addressed on the rebased head `a91f66d` (now on top of `1941e64`):

- **`indexVersionMarker` (blocking)**: `CLAUDE_CANONICAL_TRANSCRIPT_MARKER` and `CODEX_CANONICAL_TRANSCRIPT_MARKER` bumped to `_v3_` in the same commit, so discovery re-reads already-indexed sessions. Kimi untouched (whole-file reader). The one-time full re-index cost is called out in the PR description.
- **Rebase**: done onto `1941e64`; the `createRequire` → named-imports conflict in `parsing.ts` is resolved (`readSync(fd, buf, 0, READ_BUFFER_SIZE, null)` + `StringDecoder`). PR is `MERGEABLE` again.
- **Verification (re-run on the rebased head)**: `npm test` → **309 pass / 0 fail**; `typecheck` 0 errors; `lint` 0 errors (4 pre-existing warnings in untouched test files); `build:cli` / `build:core` / `electron-vite build` all pass. `test:electron:all` → 5/5 per your measurement; locally 4/5 — `electron-session-virtualization`'s final renderer-timing wait times out only under this sandboxed terminal, and this PR carries zero `app/` changes.
- **Fixture**: no more injected `padding` field — the byte offset is reached by padding `message` / `output_text` themselves, so the fixture is shaped like real Codex output (and now exercises dedup on a long assistant message).
- **Boundary constant**: `READ_BUFFER_SIZE` is exported from `parsing.ts` and the test derives the offset from it, plus a self-calibration assertion that `中`'s first byte lands on the last byte of a read chunk — the test can no longer silently degenerate if the buffer size changes.
- **Nit**: `parsing.ts` header comment now lists `node:string_decoder`.
- **CI box**: unticked — `tests/codex-parse.test.mjs` isn't in the `cli.yml` whitelist (tracked in #34).

PR #35, opened 2026-08-04
feat(core): add provider indexing foundations

SaladDay:
## Summary

This PR adds the provider-neutral indexing contracts needed by sources with non-linear histories and provider-owned snapshots. It is the base for #23 and does not include the Pi adapter.

- Store opaque provider cursors and provider-owned projection markers.
- Make session replacement and retraction atomic, including all-or-nothing force rebuilds.
- Let discovery report incomplete inventories without blocking readable units during ordinary indexing.
- Add canonical `visible`, `inactive`, and `hidden` evidence states. Standard queries remain visible-only; `includeInactive` opts into source-attested superseded evidence, while hidden evidence is never returned.
- Centralize provider root resolution for the app, worker, and passive CLI. Invalid explicit roots stay disabled instead of falling back to another corpus.
- Preserve the last verified snapshot when settings, enumeration, parsing, or publication fails.

The registry remains the only shared integration point; indexing, settings, querying, and the renderer do not branch on provider IDs.

## Compatibility

| Component | Version |
|---|---|
| Obelisk app | `0.2.2` |
| Obelisk core | `0.2.0` |
| Obelisk CLI | `0.2.2` |
| Base commit | `1941e64` |
| Validation runtime | Node `24.1.0`, npm `11.3.0`, macOS |

## Verification

- `npm test` — **342/342 passed**
- `npm run typecheck` — passed
- `npm run build:core` — passed
- `npm run build:cli` — passed
- `npm run build:skill` — passed
- `npm run lint` — 0 errors; 4 existing warnings
- `git diff --check` — passed

Focused regressions cover malformed settings, invalid roots across the desktop worker boundary, partial-inventory recovery, force-publication safety, and missing or unreadable Claude Code, Codex, and Kimi Code inventories.

PR #37, opened 2026-08-05
fix(core): apply provider unit retractions

SaladDay:
## Summary

- Apply provider-declared `IndexUnit.retractSessionIds` in shared persistence before parsing and writing the unit.
- Keep retractions, replacement records, and cursor publication atomic in the existing per-unit transaction.
- Let Pi tombstones and identity replacements use the unit contract directly, while preserving record-level deletion for current-session replay.

Follow-up to #23.

## Verification

Environment: Obelisk 0.2.0, Node.js v24.1.0, npm 11.3.0.

- Targeted persistence and Pi suites: 67/67 passed
- `npm test`: 410/410 passed
- `npm run typecheck`
- `npm run build:core`
- `npm run build:cli`
- `npm run build:skill`
- `npm run lint`: 0 errors (4 existing warnings)
- `git diff --check`

### DIFF
--- a/docs/adr/0001-parse-core-and-persist-layers.md
+++ b/docs/adr/0001-parse-core-and-persist-layers.md
@@ -30,8 +30,8 @@ binding-agnostic and does not need a per-binding implementation.
   adapter is *pure*: it emits normalized records and never touches a database.
   Discovery receives a read-only view of the provider's already-indexed session
   paths. A full-reparse adapter can therefore attach `retractSessionIds` to a
-  replacement or tombstone unit without querying SQLite itself. Retraction and
-  replacement records commit in the same unit transaction, so a failed parse
+  replacement or tombstone unit without querying SQLite itself. Unit retractions
+  and replacement records commit in the same unit transaction, so a failed parse
   preserves the last complete snapshot.
   Provider identity may therefore be richer than a wire-level ID. Pi, whose
   explicit session IDs are project-local, deterministically namespaces the
--- a/packages/core/src/persist.ts
+++ b/packages/core/src/persist.ts
@@ -86,6 +86,7 @@ function deleteSession(db: SqliteDb, sessionId: string) {
 // (also written to index_state). `db` is any SQLite handle sharing prepare/run.
 export function persist(db: SqliteDb, unit: IndexUnit, gen: Generator<TranscriptRecord, Cursor>): Cursor {
   const st = statements(db);
+  for (const sessionId of unit.retractSessionIds ?? []) deleteSession(db, sessionId);
 
   const write = (r: TranscriptRecord) => {
     switch (r.kind) {
--- a/packages/core/src/providers/pi.ts
+++ b/packages/core/src/providers/pi.ts
@@ -1331,12 +1331,8 @@ function parsePi(unit: IndexUnit): { records: TranscriptRecord[]; cursor: string
   const meta = unit.meta as PiSessionUnitMeta | undefined;
   if (meta?.kind === 'pi-tombstone') {
     return {
-      records: (unit.retractSessionIds ?? [unit.sessionId]).map((sessionId) => ({
-        kind: 'delete-session',
-        sessionId,
-      })),
-      // A zero watermark keeps recreation discoverable even if no watcher
-      // event is available, without teaching persist about provider tombstones.
+      records: [],
+      // A zero watermark keeps recreation discoverable even if no watcher event is available.
       cursor: '0:0',
     };
   }
@@ -1375,10 +1371,6 @@ function parsePi(unit: IndexUnit): { records: TranscriptRecord[]; cursor: string
   };
   return {
     records: [
-      ...(unit.retractSessionIds ?? []).map((staleSessionId) => ({
-        kind: 'delete-session' as const,
-        sessionId: staleSessionId,
-      })),
       { kind: 'delete-session', sessionId },
       session,
       ...projected.records,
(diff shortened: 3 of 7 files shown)


tommy0103 (commented):
Review follow-up covering #37 itself and the remaining findings from a full re-review of the #35 contracts (landed via #23). Verified against commit `96d8523` / base `1941e64`, with targeted suites run locally (67/67 pass for the #37 persistence + Pi tests).

## On #37 itself

The `retractSessionIds` fix looks correct and complete:

- `persist()` applies unit retractions before writing records, inside the existing per-unit / caller-owned transaction, so rollback semantics hold — the new failed-replacement test asserts this for real.
- Both the core CLI path and the app path go through the same core `persist` (via `indexProviderPlan` / `indexProviderPlanStrict`), so there is no divergent second implementation for retraction.
- Removing the record-level delete emission from Pi tombstones/identity replacements is safe: the only tombstone construction site always sets `retractSessionIds`, and retracted ids are always non-valid sessions (valid ones go through `forceSessionIds`), so no ordering hazard where one unit retracts what another re-emits.

No new issues found in #37. Good to merge from my side.

## Remaining issues introduced by #23 (commit `96d8523`), not covered by #37

**Blocking**

1. **Kimi marker-replay is a silent no-op** — `provider-indexing.ts` builds `replayKeys` from `sessions.jsonl_path` and resets cursors with `DELETE FROM index_state WHERE jsonl_path = ?`, but kimi's cursor key is the session *directory* (`kimi.ts` discovery keys units by `sessionDir`) while the session row stores `<dir>/main/wire.jsonl`. The DELETE never matches. Failure scenario: canonical-transcript marker bump + a partially unreadable kimi `sessions/` tree during replay → missed kimi sessions keep matching their stale `sessionDir` cursors and stay on old-semantics rows indefinitely. Also, `committed.has(provider\0key)` never matches for kimi, so a useless DELETE fires per kimi session per rebuild. Existing tests only use fake providers where `jsonl_path === unit.key`. The same key mismatch also affects the `raw()` cursor lookup (`query.ts`) and the app's `db:getMessageFullText` cursor plumbing, both of which key `index_state` by `sessions.jsonl_path`.

**Major**

2. **`failures()` silently drops orphaned error results** — the query changed from `LEFT JOIN messages rm` (base `1941e64`) to INNER JOINs on `messages rm`, `tool_calls tc`, `messages cm`. Codex emits `tool_result` rows with `message_uuid: ''` when a `function_call_output` has no matching call in the same file (the `|| ''` fallback in `codex.ts` exists because this shape occurs). Previously these surfaced in `failures()`; now they vanish — a failure-diagnosis helper losing exactly the broken-pair evidence.
3. **`failures()` visibility label disagrees with its admission gate** — the row's `visibility` comes from the result message, but admission also gates on the call message. A row with an `inactive` call and a `visible` result is admitted under `includeInactive: true` yet labeled `visible`, so consumers can present superseded evidence as current.
4. **Duplicate indexer-service leak in the app** — `startIndexerService` overwrites the module-level service without stopping an existing one (pre-existing), and #35 widened the race: the rebuild `finally` restarts unconditionally while the autoRefresh branch now starts on `value !== false` alone. Toggling autoRefresh (or changing a provider root) while a manual rebuild is in flight leaks a service instance for the app's lifetime — duplicate chokidar watchers, duplicate heartbeat, doubled debounced builds.

**Minor**

5. After an upgrade, queries can hit an unmigrated DB during the skip window: `openReadDb` never migrates, and `shouldSkipBuild` returns before `openDb()`, so for ~30s the new `visibility`/`cursor` column references fail with "no such column".
6. Disabled-provider stub reports `defaultRoot` (a path the user never configured) as the failing path, and reports incomplete inventory on every discover — a stale invalid root for an unused provider permanently blocks force rebuilds with `incomplete_snapshot`.
7. Orphan `tool_result` messages now fall through to the generic message branch in session detail and render as empty user bubbles, also breaking assistant-message merging.
8. Force-failure diagnostics shadowing in the app indexer (`(error as {obelisk?}).obelisk ?? error.sourceError?.obelisk` — the fallback is dead), and `attune` silently proceeds on a stale index when settings are unavailable (the query path at least warns).
9. Re-enabling autoRefresh starts the service with `buildOnStart: false`, so changes made while it was off are never caught up (the provider-root path uses `true`).
10. The new `onIndexUpdated(() => loadSettings())` subscription in `Settings.vue` clobbers a half-typed Recap path on every debounced build.
11. Main now throws on rebuild with corrupt settings, but the renderer's `rebuildIndex()` has no `catch` — the button silently resets with no error surfaced.
12. Core's force path has no direct test; it is only covered through the app's duplicated force orchestration, so divergence between the two would go unnoticed.

## Pre-existing issues (not introduced by #23, surfaced during this review)

- **The opaque-cursor contract was already self-contradictory at base**: `types.ts` declares the cursor adapter-opaque while `persist.ts` and the app indexer `split(':')` it (both present at `1941e64`). All four builtin adapters happen to keep `mtime` in the first slot — Pi's cursor comment shows it had to contort its format to comply. Either persist should stop parsing the cursor, or the contract should state the `mtime:` prefix is required. Worth deciding before a third-party adapter takes "opaque" literally.
- Saving any setting over a corrupt settings file silently replaces it with `{ <one key> }` — `loadPersistedSettings` returned `{}` on parse failure at base too.
- `subagents()` maps the after/before time filter to `sa.session_id` (`query.ts` at base), so time filters compare session ids.

PR #38, opened 2026-08-06
fix: correct provider replay and refresh behavior

SaladDay:
## Summary

- Use provider-owned unit identities for marker replay and raw cursor lookup, including Kimi's directory-backed sessions.
- Preserve orphan failure evidence and effective inactive visibility without rendering empty orphan tool-result bubbles.
- Keep destructive rebuilds, schema upgrades, and desktop refresh/rebuild behavior safe across invalid settings, incomplete inventories, and concurrent ownership.

Follow-up to #23. This PR is independent of and compatible with #37.

## Verification

Environment: Obelisk 0.2.0 (CLI 0.2.2), Node.js v24.1.0, npm 11.3.0.

- `npm test`: 423/423 passed
- Combined locally with #37: 425/425 passed
- Fixed-seed Pi 0.83.0 differential: seed `0x5eedc0de`, 512 cases, 18,124 entries, 11,586 messages, 2,230 compactions, and 1,060 orphan-parent cases
- `npm run typecheck`
- `npm run build:core`
- `npm run build:cli`
- `npm run build:skill`
- `npm run lint`: 0 errors (4 existing test warnings)
- `git diff --check`

### DIFF
--- a/app/src/main/index.ts
+++ b/app/src/main/index.ts
@@ -12,6 +12,7 @@ import { buildRecapExportQuery } from './recap-capture-query.ts';
 import { buildEditorUrl, DEFAULT_EDITOR_SCHEME, EDITOR_SCHEMES, resolveFileReference } from './file-reference.ts';
 import { acquireWriterLease, writerLockPathFor } from '../../../packages/core/src/writer-lease.ts';
 import { migrateCoreSchemaColumns } from '../../../packages/core/src/schema-migrations.ts';
+import { storedSessionCursor } from '../../../packages/core/src/provider-indexing.ts';
 import { createBuiltinProviderRegistry } from '../../../packages/core/src/providers/builtins.ts';
 import {
   createConfiguredBuiltinProviderRuntime,
@@ -269,10 +270,11 @@ function appendWhere(sql, params, clause) {
 }
 
 function startIndexerService({ buildOnStart = false } = {}) {
+  if (indexerService) return indexerService;
   const paths = getRuntimePaths();
   if (latestSettingsError !== null) return null;
   migrateLegacyDbIfNeeded(paths);
-  indexerService = createIndexerService({
+  const service = createIndexerService({
     projectsDir: paths.projectsDir,
     watchDirs: paths.providerRegistry.watchRoots(paths.providerRoots),
     buildIndex: async ({ reason, changedPaths }) => {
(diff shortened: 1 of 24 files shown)


tommy0103 (commented):
Verified this PR against the review checklist from #37's thread: all 12 issues introduced by #23 are fixed, and the fixes are real (traced the mechanisms, not just the description). Full suite passes locally (423/423), and the combination with #37 merges cleanly with 425/425.

Remaining low-severity findings from this pass — none blocking merge:

1. **Incomplete-inventory retry has no backoff (must-fix, tracked in #40)** — the new `incompleteInventory` branch in `app/src/main/indexer-service.ts` schedules a full inventory retry every `heartbeatMs` (30 s) forever if the failure is persistent (e.g. a permanently unreadable subdirectory under a provider root). Each retry is a full provider scan plus `openDb()` + `notifyIndexUpdated()`, so renderer refreshes and disk IO are spammed for the app's lifetime. Please add backoff and/or a consecutive-failure cap. Filed as #40 — must be fixed, but does not block merging this PR.
2. **`rethrowUnlessSchemaBlocked` can mask the original error** (`packages/core/src/core.ts`) — any query/sandbox error (including a bug in a user's CodeAct script) triggers a writer-lease acquisition, and if the schema is genuinely not ready the user sees `schema upgrade is blocked by writer_busy/daemon_active` instead of their original `TypeError`. Bounded impact, but the misattribution will confuse.
3. **Orphan `failures()` rows vanish under time filters** (`packages/core/src/query.ts`) — `buildWhere` binds the timestamp alias to `rm.timestamp`, which is NULL for orphan rows, so any `after`/`before`-bounded `failures()` call drops them again; they also sort last under `ORDER BY rm.timestamp DESC` and can be cut by `limit`. Inherent trade-off of preserving the evidence, but worth documenting.
4. **Disabled-provider stub now inherits the real `raw()`** (`packages/core/src/provider-settings.ts`) — after a root is de-configured, previously indexed sessions remain raw-readable from the provider's default root instead of returning null. Harmless for indexing (discover returns `[]`), but please confirm this behavior change is intended.
5. **`sessionDirectoryFromWirePath` is a layout heuristic** (`packages/core/src/providers/kimi.ts`) — a workspace directory literally named `agents` containing a session directory named `main` would be misread as canonical layout. Practically unreachable with kimi's generated IDs; a comment noting the assumption would suffice.
6. **Residual settings clobber path** (`app/src/renderer/src/views/Settings.vue`) — `browseSourcePath` still calls a full `loadSettings()`, so an uncommitted Recap path edit is still lost if the user picks a source folder via the OS dialog mid-edit. Minor since the dialog steals focus anyway.
7. Cosmetic: the rebuild error shown in Settings carries Electron's IPC prefix (`Error invoking remote method 'settings:rebuildIndex': Error: ...`).

Approving from my side — please land this together with #37 (verified compatible in both merge orders, no conflicts).

PR #48, opened 2026-08-11
fix(core): handle remember without session_id

wutongyuonce:
## Summary

- Skip the session project lookup when remember() is called without session_id.
- Store the optional project and session fields as SQLite NULL when neither is provided.
- Cover the public --attune path with remember({ path, summary }).

## Verification

- npm test: 427 passed, 0 failed.
- npm run typecheck: root and app TypeScript configurations passed.
- npm run lint: 0 errors, 4 pre-existing warnings.

## Scope

No schema, dependency, or app changes.

### DIFF
--- a/packages/core/src/query.ts
+++ b/packages/core/src/query.ts
@@ -713,7 +713,9 @@ function createAttuneApi(db: SqliteDb) {
     const normalizedPath = resolveMemoryPath(memoryPath, session_id);
     const normalizedAnchors = normalizeAnchors(anchors);
     const id = `mem-${Date.now()}-${Math.random().toString(36).slice(2, 8)}`;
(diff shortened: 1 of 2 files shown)


tommy0103 (dismissed):
(Retracted — this review was posted prematurely before maintainer discussion concluded. Please disregard; a consolidated response will follow.)

tommy0103:
Thanks for the PR — the underlying issue you found is real (`remember()` without `session_id` dies with an opaque `ERR_INVALID_ARG_TYPE` rather than a deliberate behavior).

That said, I'm going to close this for now: the memory layer (`--attune` / `remember()`) is still an experimental feature, and some of its architecture decisions aren't settled yet — including whether `session_id` should be required rather than optional. Until that direction is decided, I'd rather not take a behavior change in either direction.

Appreciate the investigation either way.

PR #49, opened 2026-08-11
fix(claude): link workflows by unique run id

wutongyuonce:
## What and why

When a Claude session runs the same workflow name more than once, the workflow-name fallback can attach the newer workflow record to an older Workflow tool call. This change links workflows only when the tool result contains the unique run ID; workflowName remains available as display metadata.

The regression test exercises the Claude provider through discover() and parse() with two same-name runs and verifies that the newer run links to new-call.

## Verification

- [x] npm test — 450 passed, 0 failed.
- [x] npm run typecheck — 0 errors across root and app configurations.
- [x] npm run lint — 0 errors, 4 pre-existing warnings.
- [x] node --test tests/claude-parse.test.mjs — 5 passed, 0 failed.
- [x] No existing assertion was loosened.
- [x] Re-ran all checks after rebasing onto upstream/main at f8cdb61.

## Provider adapter checks

- [x] Read claude.ts, codex.ts, kimi.ts, ADR-0001, and ADR-0007.
- [x] Regression test calls discover() against the existing Claude workflow directory layout.
- [x] Canonical transcript and SQLite round-trip suites pass.
- [x] No identity, record-shape, visibility, cursor, schema, or version-marker change.

## Deliberately out of scope

No CI workflow change. The repository's current CLI workflow does not include claude-parse.test.mjs.

No schema migration or index-version bump: this corrects selection of an existing parent_tool_use_id field without changing stored row shape.

### DIFF
--- a/packages/core/src/providers/claude.ts
+++ b/packages/core/src/providers/claude.ts
@@ -180,7 +180,6 @@ function toolResultText(content: unknown): string {
 function workflowParentToolUseId(
   transcriptPath: string,
   runId: string,
-  workflowName: string | null,
 ): string | null {
   if (!existsSync(transcriptPath)) return null;
   const workflowToolIds = new Set<string>();
@@ -202,7 +201,7 @@ function workflowParentToolUseId(
     for (const block of content) {
       if (block?.type !== 'tool_result' || !workflowToolIds.has(block.tool_use_id)) continue;
       const text = toolResultText(block.content);
-      if (!text.includes(runId) && !(workflowName && text.includes(workflowName))) continue;
+      if (!text.includes(runId)) continue;
       parentToolUseId = block.tool_use_id;
       return false;
     }
@@ -226,7 +225,6 @@ function* parseWorkflow(unit: IndexUnit): Generator<TranscriptRecord, Cursor> {
     parent_tool_use_id: workflowParentToolUseId(
       meta.mainTranscriptPath,
       workflow.runId,
-      workflow.workflowName || null,
     ),
     task_id: workflow.taskId || null,
     script: workflow.script || null,
--- a/tests/claude-parse.test.mjs
+++ b/tests/claude-parse.test.mjs
@@ -162,3 +162,38 @@ test('claude provider emits workflow artifacts with an explicit canonical tool e
   assert.deepEqual(persistedDetail, detail);
   db.close();
 });
+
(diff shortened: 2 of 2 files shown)


tommy0103 (changes requested):
The fix itself looks right to me: matching the Workflow tool call by unique run id only (dropping the workflowName fallback) fixes the mis-linking when the same workflow name runs more than once, and the repo's own golden fixture ('run-workflow complete' — contains the run id, not the workflow name) supports runId being the reliable token. Worst case for older transcripts is an honest null parent_tool_use_id instead of a wrong link. The regression test design is good too.

One blocking issue, though: this PR's base predates #51, and merging it as-is breaks the test suite on main. After #51, `tests/claude-parse.test.mjs` uses `makeTempDir` from `./temp-dirs.mjs` (tests must clean up their temp dirs on process exit) and no longer imports `mkdtempSync`/`tmpdir`. The textual merge is clean (GitHub shows MERGEABLE), but the merged file's new test references `mkdtempSync` and `tmpdir` without importing them — I verified with `git merge-tree` + `node --test tests/claude-parse.test.mjs` that the merged result fails with a ReferenceError at the new test. And as you noted, the CLI workflow doesn't run this test file, so CI wouldn't catch it.

Could you rebase onto current main and switch the new test to `makeTempDir('obelisk-claude-workflow-link-')` (dropping the `mkdtempSync`/`tmpdir` imports)? Happy to merge right after that.

wutongyuonce:
Rebased onto current `upstream/main` (`f8cdb61`) and updated the new regression test to use `makeTempDir('obelisk-claude-workflow-link-')`.

Verification on the rebased head:

- `node --test tests/claude-parse.test.mjs`: 5 passed, 0 failed
- `npm test`: 450 passed, 0 failed
- `npm run typecheck`: 0 errors
- `npm run lint`: 0 errors, 4 pre-existing warnings

tommy0103:
Closing. This was already on our own fix list, and #52 now covers it as part of a more comprehensive fix for workflow-to-tool-call linking (run-id-only matching plus finalize-time healing of links the file-scan path misses). We won't be merging this PR.

tommy0103:
For the avoidance of doubt: this closure is not an absorption. The equivalent fix was on our own fix list before this PR was opened, and #52 would have landed regardless. This PR is simply not taken.

PR #65, opened 2026-08-16
feat(codex): index archived sessions

luoling8192:
## What and why

Adds `~/.codex/archived_sessions` as a Codex transcript root.

Codex archive JSONL files were previously unreachable because the provider only
discovered, watched, and searched `~/.codex/sessions`. This change includes the
archive directory in all three paths while preserving existing session IDs and
canonical record shapes.

- discovers archived JSONL files during full and changed-path indexing;
- watches archive changes in the desktop indexer; and
- resolves archived raw records, including child-thread lookups.

## Verification

- [x] `npm test` — 383 pass / 0 fail
- [x] `npm run typecheck` — 0 errors
- [x] `npm run lint` — 0 errors; 4 pre-existing warnings
- [x] New provider and runtime tests run through the standard Node test suite
- [x] Provider discovery, watch-root registration, raw lookup, and runtime
      indexing were exercised end to end

## Deliberately out of scope

No schema, session-identity, or canonical-transcript changes. This only makes
Codex's existing archived JSONL corpus available through the current provider.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ The agent writes JS queries, runs them locally, and answers in plain language.
 
 **App side** — an Electron desktop app for humans to browse sessions, manage memories, view usage stats, and see weekly recap cards.
 
-Both read from the same `~/.obelisk/obelisk.sqlite` database. The indexer reads Claude Code transcripts from `~/.claude/projects`, Codex transcripts from `~/.codex/sessions`, Kimi Code sessions from `~/.kimi-code/sessions` (or `$KIMI_CODE_HOME/sessions`), and Pi sessions from `~/.pi/agent/sessions`.
+Both read from the same `~/.obelisk/obelisk.sqlite` database. The indexer reads Claude Code transcripts from `~/.claude/projects`, Codex transcripts from `~/.codex/sessions` and `~/.codex/archived_sessions`, Kimi Code sessions from `~/.kimi-code/sessions` (or `$KIMI_CODE_HOME/sessions`), and Pi sessions from `~/.pi/agent/sessions`.
 
 ## Multi-provider support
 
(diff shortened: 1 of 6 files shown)


tommy0103:
One behavior change worth calling out explicitly, since the PR body doesn't mention it (and the "no canonical-transcript changes" framing could read as "no behavior change on the existing sessions path"):

In `discoverAt`, the cursor-skip guard gained a `!fileChanged` condition:

```ts
if (!sessionIndexChanged && !fileChanged && cursor !== null && Number(cursor.split(':')[0]) >= statSync(file.path).mtimeMs && guardian === null) {
```

**Effect:** in changed-paths (watch/daemon) mode, a file reported by the watcher is now always re-parsed, even when the stored cursor mtime is `>=` the file's mtime. Previously such no-op events were skipped. Full-scan mode is unaffected — `changedFiles` is empty there, so the guard behaves exactly as before.

I believe this change is correct and worth keeping:

- The mtime guard remains indispensable in full-scan mode, where it is the only change signal — that path is untouched.
- In changed-paths mode the watcher is the stronger signal; the mtime check could only veto it, leaving a permanently stale index when content changed without mtime advancing (same-ms rewrite, file replaced by an older-mtime copy). The next full inventory would not heal that either, since it applies the same mtime check.
- It aligns codex with the existing convention in pi (`pi.ts` skips via cursor only when `changedPaths === null`) and kimi, which also trust watcher-reported paths despite being full-reparse providers as well.
- The cost is bounded: the indexer debounces and dedups watcher events, and during active sessions the file is genuinely growing (mtime advancing), so the guard was rarely saving work anyway.

Suggest adding a line to the PR body so this is visible to future archaeology, e.g.: "Changed-paths mode now trusts watcher-reported paths over the mtime cursor, aligning codex with the pi/kimi providers."

PR #69, opened 2026-08-17
Keep Kimi undo replay in sync

### LINKED ISSUE #26: [Providers] Kimi undo replay keeps injection messages that Kimi itself drops
## TL;DR

Replaying a Kimi `context.undo` event leaves injection messages in the index that Kimi itself removed from the context. Low impact today — the leftovers are reminder-class records — but the divergence should be closed before undone content is ever surfaced rather than dropped.

> **Priority: low.** No user-visible symptom has been observed. This is a fidelity gap found by reading both implementations side by side, not a reported bug.

## Problem

Kimi's `context.undo` is a tail truncation at a user-turn boundary: it removes the last N *real user prompts* and everything that followed them. The Kimi adapter replays that event to keep the index consistent with what Kimi actually has in context.

The two implementations disagree on one class of record. Kimi counts injection messages as non-undoable *when deciding where to cut*, but still removes them *when applying the cut*. The adapter treats "skip while counting" as "skip while removing", so injections inside the undone range survive into the index.

## Current behavior

`packages/core/src/providers/kimi.ts:294-309` walks backwards from the tail and splices each message until it has removed `count` real user messages, but `continue`s past any message whose `origin.kind === 'injection'` (tracked at `kimi.ts:385`), leaving it in place:

```ts
for (let index = messages.length - 1; index >= undoFloor; index--) {
  const message = messages[index]!;
  if (injectionMessageUuids.has(message.uuid)) continue;   // left in the projection
  messages.splice(index, 1);
  ...
}
```

Kimi's own operation resolves a single cut index and slices, so nothing after that index survives regardless of origin. For reference, in the `kimi-code` repository: `packages/agent-core-v2/src/agent/contextMemory/contextOps.ts` — `computeUndoCut()` skips injections only while counting real user prompts to locate `cutIndex`, and `contextUndo` then applies `state.slice(0, cutIndex)`.

Net effect: for an undone range that contained injection messages, Obelisk's projection is a superset of Kimi's context.

## Expected behavior

Replaying one `context.undo` event yields the same surviving message set as Kimi's own context after that undo.

## Scope

Module: Kimi provider undo replay.

| File | Lines | Current state |
|------|-------|---------------|
| `packages/core/src/providers/kimi.ts` | 294-309 | `applyUndo` preserves injection-origin messages while removing everything else in the undone range |
| `packages/core/src/providers/kimi.ts` | 385 | Populates `injectionMessageUuids` from `origin.kind === 'injection'` |

Upstream reference for comparison (external repo, not a dependency): `kimi-code` → `packages/agent-core-v2/src/agent/contextMemory/contextOps.ts`.

## Reproducing

Not reproduced from captured data — across 19 local `wire.jsonl` files there was not a single `context.undo` record, so a fixture has to be constructed. A wire log shaped like the following is enough:

1. `context.append_message` — a real user prompt (`origin.kind === 'user'`)
2. `context.append_message` — an assistant reply
3. `context.append_message` — an injected message (`origin.kind === 'injection'`)
4. `context.undo` with `count: 1`

Index that session and read back its messages: the injected record from step 3 is still present.

Ways a real `context.undo` gets written, if capturing one from a live session is preferred over a synthetic fixture: `/undo [count]` in the TUI, the interactive undo selector (`/undo` with no argument), `/undo` in the web app, and — most commonly — editing and resending the last user message in the web app, which issues `undo(1)` first.

Worth noting while writing the fixture: Kimi refuses the whole operation when the cut would cross a compaction summary, when there is nothing to undo, or when fewer than `count` turns are undoable. It is all-or-nothing, and the precheck runs before the wire event is written, so a `context.undo` present in a log is always one that fully applied.

## Acceptance criteria

- [ ] For a wire log whose undone range contains injection-origin messages, the adapter's surviving message set matches Kimi's `slice(0, cutIndex)` semantics
- [ ] A fixture test covers that case
- [ ] `npm test` passes

## Related

Relevant to the visibility model under discussion in #23. If retracted Kimi records are ever kept and marked instead of deleted, this divergence would split them — part marked, part incorrectly left as ordinary visible history — so it is worth closing first or at the same time.

yhuikzdtguioaert:
Fixes #26.

## Reproduction and root cause

A Kimi transcript with a real user message followed by `origin.kind = "injection"` and `context.undo(count: 1)` kept the injection in the canonical transcript. `applyUndo` skipped injection messages entirely while walking backward, although Kimi drops every message inside the undone range and only excludes injections from the user-message count.

## Changes

- remove every message encountered while replaying the undone range, while continuing to count only real user messages toward `count`
- correct the existing undo fixture so its assertion describes Kimi's surviving message set
- bump the Kimi canonical transcript marker so previously indexed sessions are replayed once

## Verification

- regression assertion against the original implementation: fails with the unexpected `persistent injection`
- `node --experimental-test-module-mocks --test tests/kimi-parse.test.mjs tests/app-kimi-index.test.mjs` — 14 passed
- `npm run typecheck` — passed
- `npm run lint` — passed (4 pre-existing warnings, 0 errors)

### DIFF
--- a/packages/core/src/providers/kimi.ts
+++ b/packages/core/src/providers/kimi.ts
@@ -60,7 +60,7 @@ interface ProjectedSession {
 }
 
 const SOURCE = 'kimi';
-export const KIMI_CANONICAL_TRANSCRIPT_MARKER = '__kimi_canonical_transcript_v4__';
+export const KIMI_CANONICAL_TRANSCRIPT_MARKER = '__kimi_canonical_transcript_v5__';
 
 function defaultKimiRoot(): string {
   return process.env['KIMI_CODE_HOME'] ?? join(homedir(), '.kimi-code');
@@ -301,7 +301,6 @@ function projectSession(meta: KimiSessionUnitMeta, sessionId: string, state: Jso
       let removedUserCount = 0;
       for (let index = messages.length - 1; index >= undoFloor; index--) {
         const message = messages[index]!;
-        if (injectionMessageUuids.has(message.uuid)) continue;
         messages.splice(index, 1);
         removedMessageUuids.add(message.uuid);
         injectionMessageUuids.delete(message.uuid);
--- a/packages/core/src/providers/kimi.ts
+++ b/packages/core/src/providers/kimi.ts
@@ -60,7 +60,7 @@ interface ProjectedSession {
 }
 
 const SOURCE = 'kimi';
-export const KIMI_CANONICAL_TRANSCRIPT_MARKER = '__kimi_canonical_transcript_v5__';
+export const KIMI_CANONICAL_TRANSCRIPT_MARKER = '__kimi_canonical_transcript_v6__';
 
 function defaultKimiRoot(): string {
   return process.env['KIMI_CODE_HOME'] ?? join(homedir(), '.kimi-code');
@@ -285,8 +285,8 @@ function projectSession(meta: KimiSessionUnitMeta, sessionId: string, state: Jso
     const stepStarts = new Map<string, number>();
     const stepMessages = new Map<string, MessageRecord[]>();
     const callMessageUuids = new Map<string, string>();
-    const injectionMessageUuids = new Set<string>();
-    const realUserMessageUuids = new Set<string>();
+    const injectionOwnerPromptIds = new Map<string, unknown>();
+    const realUserPromptIds = new Map<string, unknown>();
     let undoFloor = wireMessageStart;
 
     const resetOpenState = (): void => {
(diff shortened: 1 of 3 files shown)


tommy0103 (commented):
Verified the fix against the upstream `kimi-code` implementation (`packages/agent-core-v2/src/agent/contextMemory/contextOps.ts` @ 698b6468b). The counting semantics match exactly — upstream `isUndoAnchor` (`conversationTime.ts:13-21`) is equivalent to the adapter's `isRealUserMessage`, so `system_trigger` messages correctly don't count toward `count`. Tests pass locally (14/14), and the new assertion fails against the old implementation as claimed.

Two findings:

**1. The removal range is still incomplete — missing the `isPromptOwnedInjection` pullback.**

Upstream `computeUndoCut` doesn't stop at the anchor; it keeps walking back past injections owned by the undone prompt:

```ts
cutIndex = i;
while (cutIndex > 0 && isPromptOwnedInjection(state[cutIndex - 1]!, message)) {
  cutIndex--;
}
```

This is a real ordering, not a theoretical one. `promptService.ts:243-254` (`appendPrompt`) appends image-compression caption reminders — `origin: {kind: 'injection', variant: 'image_compression', ownerPromptId: X}` via `systemReminderService.ts:29` — *before* appending the user prompt with `id: X`. So the wire log looks like:

```
context.append_message  injection(ownerPromptId=X)   // precedes the prompt
context.append_message  user prompt(id=X)
context.undo(count=1)
```

The new `applyUndo` breaks immediately after splicing the `count`-th real user message, so those owned injections survive in the index while Kimi removes them. Against this issue's own acceptance criterion ("surviving message set matches Kimi's `slice(0, cutIndex)` semantics"), sessions containing image-compression captions still diverge. The fixture's `persistent injection` has no `ownerPromptId` and sits *after* the prompt, so it doesn't cover this ordering.

Suggested direction: track prompt `id` → uuid during projection, and after the undo loop reaches its anchor, keep splicing backward while the preceding message is an injection whose `ownerPromptId` equals that anchor's id. Same low severity as the original issue (reminder-class leftovers), but worth closing in this PR or at least noting as a known follow-up.

**2. Dead state: `injectionMessageUuids` is now write-only.**

After removing the `continue` guard, the set is only added to (`kimi.ts:388`) and deleted from (`kimi.ts:306`) but never read. The declaration and both operations can go.

yhuikzdtguioaert:
Undo now removes contiguous prompt-owned injections preceding the undone prompt, and the unused `injectionMessageUuids` state is removed. Added regression coverage; targeted tests, typecheck, and lint pass.

tommy0103 (approved):
The follow-up commit resolves both findings from my earlier review. I verified the pullback logic against upstream `computeUndoCut` branch by branch — including count > 1 (only the earliest anchor's pullback determines the final cut), prompts without an `id` (no pullback, matching `isPromptOwnedInjection`), and non-contiguous owned injections (both sides stop) — the surviving message sets match in every case. The new fixture test fails against the previous implementation and passes here; full suite 465/465, typecheck and lint clean; dead `injectionMessageUuids` state is gone. Marker v5→v6 bump is correct since the projection changed again.

One optional nit, not blocking: `injectionOwnerPromptIds` and `realUserPromptIds` are typed `Map<string, unknown>`. Since the stored values are `origin.ownerPromptId` / `source.id` (string or absent), `Map<string, string | undefined>` would be tighter and would let the compiler check the `anchorPromptId` comparison instead of relying on the `typeof` guard alone.

PR #77, opened 2026-08-19
perf(core): stop rebuilding the whole messages_fts index in every incremental finalize

### LINKED ISSUE #75: Incremental builds rebuild the whole messages_fts index every finalize — ~30s fixed cost independent of change size
## What happens

Every build — incremental or not — runs `INSERT INTO messages_fts(messages_fts) VALUES('rebuild')` in its finalize transaction (`packages/core/src/indexer.ts`). The rebuild regenerates the full-text index from the entire `messages` table, so its cost is proportional to the index size, not to how much changed.

Measured on a real-world index (1.8 GB, ~688k messages, node:sqlite / SQLite 3.54):

| Step | Wall time |
|---|---|
| `messages_fts('rebuild')`, unicode61 | ~10.3 s |
| same, trigram tokenizer (configurable per #9/#14) | ~30.2 s |
| stat-compare of 8,669 jsonl files | 0.07 s |

For a query-path freshness build (the invocation-nonce recovery) or any frequent refresh, this rebuild is the dominant fixed cost — an otherwise ~zero-change incremental build pays 10–30 s every time.

## Why the rebuild appears to be redundant on the incremental path

`messages_fts` is an external-content table maintained by the `messages_fts_ai/au/ad` triggers, and the persist layer writes messages with `INSERT ... ON CONFLICT(uuid) DO UPDATE` (fires AU) and deletes them with plain `DELETE` (fires AD, including the force path's `DELETE FROM messages` and `deleteSession`). No current write shape bypasses the triggers — the classic hazard (`INSERT OR REPLACE`, whose implicit delete fires no DELETE trigger without `recursive_triggers`) does not exist in the messages write path today. An FTS5 `integrity-check` (rank=1) passes against the live 688k-message index.

## Proposal

Run the wholesale rebuild exactly once per index — to heal databases whose rows predate trigger-only maintenance — recorded under a marker in `index_state`, written in the same transaction as the rebuild (interrupted build redoes it). The force path keeps its wholesale rebuild as the last line of defence and re-records the marker it wipes.

I have this implemented with tests (poison-posting probe proving the rebuild is skipped incrementally, still performed on force, and performed exactly once for a marker-less index; upsert/delete trigger-semantics pins; FTS5 integrity-check after each path) and can open the PR right away if the direction sounds right. Happy to adjust if you prefer a different invalidation strategy.

Measured effect on this machine's daemon-style refresh loop: incremental build 40–56 s → 10.6–11.4 s with real changes, 2.4 s with none.

Cyb3rN8:
Fixes #75.

Every finalize runs `messages_fts('rebuild')`, regenerating the FTS index from the whole `messages` table. Cost is proportional to index size, not change size: measured on a 1.8 GB / 688k-message index, ~10.3 s under unicode61 and ~30.2 s under trigram (#14), out of a ~50 s incremental build. File stat-comparison (8,669 jsonl, 0.07 s) is not the bottleneck — the rebuild is.

The rebuild is redundant on the incremental path: `messages_fts` is external-content maintained by the `ai/au/ad` triggers, persist upserts messages via `ON CONFLICT DO UPDATE` (fires AU) and deletes via plain `DELETE` (fires AD). No current write shape bypasses the triggers; FTS5 `integrity-check` (rank=1) passes on the live index.

## Change

- Incremental finalize performs the rebuild **exactly once per index** (heals rows that predate trigger-only maintenance, e.g. written by an `INSERT OR REPLACE`-era version) and records `__messages_fts_synced__` in `index_state` **in the same transaction** — an interrupted build redoes the heal; the marker never runs ahead of the work (converges when re-run, per the contributing guide's destructive-work rule).
- The force path keeps its wholesale rebuild as the last line of defence and re-records the marker it wipes with `DELETE FROM index_state`.
- `rebuildMemoryFts` is deliberately untouched: `memories` is tiny, and `--attune` writes it outside build transactions, so the cheap belt-and-suspenders refresh stays.

## Verification

- New `tests/incremental-fts-sync-marker.test.mjs`:
  - poison-posting probe (an FTS row with no `messages` counterpart): survives an incremental build (rebuild skipped), cleared by a force build, cleared exactly once for a marker-less index (the heal), with `integrity-check` after each path;
  - pins the trigger semantics for persist's exact upsert/delete shapes — the writes the wholesale rebuild used to bail out.
- Full suite on this branch head: **469/469 pass, typecheck clean, lint 0 errors** (4 pre-existing warnings in unrelated app tests, present on main).
- Real-machine effect (daemon-style refresh loop): incremental build 40–56 s → 10.6–11.4 s with real changes, 2.4 s floor with none; search behavior verified unchanged against LIKE ground truth.

### DIFF
--- a/packages/core/src/indexer.ts
+++ b/packages/core/src/indexer.ts
@@ -108,6 +108,33 @@ function healWorkflowParentLinks(db: SqliteDb): void {
 const BUILD_DEBOUNCE_MS = 30000;
 const APP_HEARTBEAT_FRESH_MS = 60000;
 
+// messages_fts is maintained row-by-row by its schema triggers: persist writes
+// messages with INSERT ... ON CONFLICT DO UPDATE (fires messages_fts_au) and
+// deletes them with plain DELETE (fires messages_fts_ad), so an incremental
+// build never leaves the index stale. The wholesale rebuild that used to run in
+// every finalize predates that guarantee, and its cost is independent of how
+// much changed — on a 1.8 GB index it is ~30 s of a ~50 s incremental build
+// (3× that under a trigram tokenizer). It is still needed exactly once, to heal
+// an index whose rows were written before trigger-only maintenance could be
+// trusted (e.g. by a version that wrote messages with INSERT OR REPLACE, whose
+// implicit deletes fire no DELETE trigger). This marker records that the heal
+// has happened; it is written in the same transaction as the rebuild, so an
+// interrupted build simply redoes it — the marker never runs ahead of the work.
+const MESSAGES_FTS_SYNC_MARKER = '__messages_fts_synced__';
+
+function markMessagesFtsSynced(db: SqliteDb): void {
+  db.prepare(
+    "INSERT OR REPLACE INTO index_state (jsonl_path, mtime, lines_processed) VALUES (?, ?, 0)",
+  ).run(MESSAGES_FTS_SYNC_MARKER, Date.now());
+}
+
+function syncMessagesFtsOnce(db: SqliteDb): void {
+  const done = db.prepare('SELECT jsonl_path FROM index_state WHERE jsonl_path = ?').get(MESSAGES_FTS_SYNC_MARKER);
+  if (done) return;
+  db.exec("INSERT INTO messages_fts(messages_fts) VALUES('rebuild')");
+  markMessagesFtsSynced(db);
+}
+
 function shouldSkipBuild(db: NodeSqliteDb, { now = Date.now(), ignoreRecentBuild = false, ignoreDaemonOwnership = false }: BuildCheckOptions = {}) {
   if (!ignoreDaemonOwnership) {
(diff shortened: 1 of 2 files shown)


tommy0103:
Superseded by #126 ("[Core] Bound ordinary incremental index finalization"), which has landed on `main`.

Latest main already implements the same fix this PR proposed: incremental finalize now calls `ensureFtsReady(db)` (`packages/core/src/indexer.ts`), which performs the wholesale `messages_fts('rebuild')` exactly once per index and records `__fts_triggers_ready__` in `index_state` in the same transaction; subsequent incremental builds skip the rebuild. The force path still passes `{ force: true }` and keeps the full rebuild as the last line of defence.

One deliberate difference from this PR: main also skips the `memories_fts` rebuild after the marker (this PR kept it), which is safe because `memories_fts` is likewise maintained by its own triggers.

Verified against `origin/main` (commit `f256668`): the per-finalize rebuild this PR set out to remove no longer happens. Closing — thanks for the original diagnosis and measurements, they pointed at the right root cause.

PR #78, opened 2026-08-19
feat(core): guard trigram sub-3-codepoint search terms; hit-centered snippets

### LINKED ISSUE #76: Trigram tokenizer: sub-3-codepoint search terms are silently dropped (false positives in multi-term queries)
## What happens

Follow-up to #9/#14 (configurable tokenizer). Under `tokenize='trigram'`, a query term shorter than three code points emits no tokens. Two failure modes:

1. **Alone it matches nothing** — `MATCH 'ok'` returns 0 rows (known, documented).
2. **Next to longer terms it silently matches everything** — the short term becomes an empty phrase that constrains nothing. Measured on a real 688k-message trigram index:

```
MATCH 'quick'     -> 799 rows
MATCH 'quick ok'  -> 799 rows   (LIKE ground truth for quick AND ok: 442)
```

357 of those hits do not contain `ok` at all. This is worse than the zero-hit case because nothing looks wrong — `search('修复 ok')`-style queries (two-code-point CJK words hit the same wall) return confidently wrong result sets. For CJK users the trigram tokenizer is the whole point, so this bites exactly the audience #14 serves.

## Proposal (implemented, PR ready)

In `search()`, detect the live tokenizer from `sqlite_master` (not from configuration, so it stays correct across migrations), and for plain token-and-whitespace queries only:

- terms ≥3 code points go through MATCH as before (rank preserved); shorter terms are pushed down as `AND m.text LIKE ?` substring conditions on the same statement — filtering the full MATCH set before LIMIT, reported via a `degraded: 'short-token-post-filter'` field on each hit;
- when every term is short, fall back to a LIKE scan of the content table (`degraded: 'like-scan'`, `rank: null`, recency-ordered) instead of returning zero rows;
- queries using explicit FTS5 operators/syntax bypass the guard; the existing punctuation fallback (catch → per-token quoting) routes through the same guard since its extracted tokens can be short too.

On the default unicode61 index the guard never engages and behavior is unchanged.

The same PR adds a hit-centered `snippet` field to every search hit (window around the earliest literal term occurrence): long transcript messages routinely match far past the head, so head-truncating previews never show the match. Happy to split that into its own PR if you prefer.

Cyb3rN8:
Fixes #76 and #194. Follow-up to #9/#14 (configurable tokenizer).

Under trigram, a term shorter than three code points emits no tokens. Alone it can only match nothing; next to longer terms it becomes an empty phrase constraining nothing. Real-index measurement: `MATCH 'quick ok'` → 799 rows, LIKE ground truth 442 — 357 silent false positives. Two-code-point CJK words (`修复`) hit the same wall, i.e. exactly the audience the trigram option serves.

## Change

`search()` detects the live tokenizer from `sqlite_master` (correct across migrations, no config dependency) and, for plain token-and-whitespace queries:

- terms ≥3 code points MATCH as before; shorter terms ride the same statement as `AND m.text LIKE ?` substring conditions, filtering the full MATCH set **before LIMIT** (rank preserved) — reported per-hit as `degraded: 'short-token-post-filter'`. An earlier draft post-filtered an overselected window in JS and collapsed to zero recall on the real index (top bm25 hits for the long term lacked the short one); the SQL push-down is recall-exact and is pinned by a buried-hit regression test.
- all-short queries fall back to a LIKE scan (`degraded: 'like-scan'`, `rank: null`, recency-ordered) instead of returning zero rows;
- the existing punctuation fallback (catch → per-token quoting) routes through the same guard.

On the default unicode61 index the guard never engages; behavior is byte-for-byte unchanged (pinned by test).

Every hit also gains `snippet` — a window centered on the earliest literal term occurrence (long transcript messages routinely match far past the head; head-truncating previews never show the match). Omitted when no term appears literally.

## Raw FTS5 syntax (#194, added 2026-09-21)

The first revision of this PR let explicit FTS5 syntax bypass the guard entirely. That was wrong, and it left the bug in place for the shape most likely to hit it: `search('"quick" ok')` returned rows without `ok` and reported `degraded: null`. Quoting one term was enough. The gate conflated "is this raw syntax" (may the text be rewritten?) with "does this query hold unindexable terms" (is a guard needed?) — the second is true either way.

Raw queries now take the same substring push-down, but only where it means the same thing:

- conjunctive raw query with at least one indexable term → short terms required as substrings, `degraded: 'short-token-post-filter'`, raw MATCH text untouched so ranking survives;
- `OR` / `NOT` / `NEAR` present, or no term long enough to MATCH → an added AND would answer a different question, so the rows stay exactly as FTS5 answered and the hit carries a new `degraded: 'short-token-unguarded'`.

FTS5 operator keywords are excluded from short-term detection, so `'"quick" OR "fine"'` is still untouched and unmarked (pinned by the existing test).

**Known limitation, documented rather than fixed:** an empty result carries no marker at all, because `degraded` rides on each hit — a query holding a short term that returns nothing may have dropped the term rather than missed it. Fixing that means changing the return shape, which felt like your call rather than mine; see #194.

## Existing-assertion change (own section per the contributing guide)

`contract-helper-shapes.test.mjs` `exactKeys` for the search() hit gains `'snippet'` — extended, not loosened: the fixture query guarantees a literal term hit, so the key is deterministically present. `api-reference.md` documents every `degraded` value and both new optional fields in the same commits (the doc-sync guard enforces this). No assertion was weakened for the #194 follow-up; the raw-syntax bypass test still asserts the operator query is untouched.

## Verification

Re-run on the current head after merging `upstream/main` (this branch was 54 commits behind; the merge conflicted only in `query.ts`, where #139's bounded temporal context landed beside the snippet-terms line — both sides kept).

- `tests/search-short-token-guard.test.mjs` now 11 tests: the original 8 plus raw-syntax filtering, raw-syntax unenforceable reporting, and unicode61-untouched-under-raw-syntax.
- Full suite: **768 tests, 746 pass, 22 fail — all 22 are `tests/kimi-manifest.test.mjs`**, failing identically on this branch with the change reverted. They are an environment incompatibility, not a regression: Node v26.4.0 with `--experimental-test-module-mocks` throws `TypeError: Cannot redefine property: constants` in `cjsMockModuleLoad` before any assertion runs. Please read the CI result as authoritative over my local numbers for that file.
- `npm run typecheck` clean (root and app tsconfig). `npm run lint` 0 errors, 11 warnings — all pre-existing in unrelated test files.
- `npm run test:electron:all` **not run**: this machine does not use the Electron app. No `app/` source is touched by either commit.
- Real-index spot check after deploy (880k messages, trigram): `search('"quick" ok')` now returns only rows containing both terms with numeric rank and `degraded: 'short-token-post-filter'`; `search('quick OR ok')` returns the disjunction unchanged with `degraded: 'short-token-unguarded'`.

### DIFF
--- a/packages/core/src/query.ts
+++ b/packages/core/src/query.ts
@@ -143,6 +143,61 @@ function buildSafeFtsQuery(text: unknown): string {
     .join(' ');
 }
 
+function queryTokens(text: unknown): string[] {
+  return String(text || '').match(/[\p{Letter}\p{Number}]+/gu) || [];
+}
+
+const FTS5_OPERATOR_TOKENS = new Set(['AND', 'OR', 'NOT', 'NEAR']);
+
+// A plain query is tokens and whitespace only — no phrases, parens, column
+// filters, or prefix stars — and none of its tokens is an FTS5 operator.
+// Anything else is honored as raw FTS5 syntax and left untouched.
+function isPlainQuery(text: unknown): boolean {
+  if (/[^\p{Letter}\p{Number}\s]/u.test(String(text || ''))) return false;
+  return !queryTokens(text).some((token) => FTS5_OPERATOR_TOKENS.has(token));
+}
+
+// The trigram tokenizer emits no tokens for terms shorter than three code
+// points. A lone short term can then only match nothing, and a short term
+// next to longer ones becomes an empty phrase that constrains nothing — the
+// query silently returns rows that do not contain the term at all. The guard
+// below only engages when messages_fts is actually built with trigram, which
+// is read from the live schema rather than configuration so it stays correct
+// across tokenizer migrations.
+function isTrigramMessagesFts(db: SqliteDb): boolean {
+  try {
+    const row = db.prepare("SELECT sql FROM sqlite_master WHERE type='table' AND name='messages_fts'").get();
+    return /tokenize\s*=\s*['"]*trigram/i.test(String(row?.sql ?? ''));
+  } catch {
+    return false;
+  }
+}
+
+const SNIPPET_BEFORE = 60;
+const SNIPPET_AFTER = 100;
+
+// Hit-centered snippet: the head of a long message often does not contain the
+// match at all, so slice a window around the earliest term occurrence instead
+// of truncating from the front. Returns null when no term appears literally
+// (e.g. raw FTS5 syntax queries), in which case the caller has message.text.
+function makeSnippet(rawText: unknown, terms: string[]): string | null {
+  const text = String(rawText ?? '');
+  if (!text || terms.length === 0) return null;
+  const hay = text.toLowerCase();
+  let hit = -1;
+  for (const term of terms) {
+    const idx = hay.indexOf(term.toLowerCase());
+    if (idx !== -1 && (hit === -1 || idx < hit)) hit = idx;
+  }
+  if (hit === -1) return null;
+  let start = Math.max(0, hit - SNIPPET_BEFORE);
+  let end = Math.min(text.length, hit + SNIPPET_AFTER);
+  // Never split a surrogate pair at a window edge.
+  if (start > 0 && (text.charCodeAt(start) & 0xFC00) === 0xDC00) start--;
+  if (end < text.length && (text.charCodeAt(end) & 0xFC00) === 0xDC00) end++;
+  return (start > 0 ? '…' : '') + text.slice(start, end) + (end < text.length ? '…' : '');
+}
+
 function createQueryApi(
   db: SqliteDb,
   {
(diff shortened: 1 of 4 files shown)


Cyb3rN8:
Updated this PR rather than opening a second one, since the bug it fixes is in this branch's own guard and would not apply to `main`.

Two things landed:

1. Merged `upstream/main` (the branch was 54 commits behind). Only `query.ts` conflicted — #139's bounded temporal context landed beside the snippet-terms line; both sides kept.
2. The #194 fix: the guard no longer skips raw FTS5 syntax. `search('"quick" ok')` used to return rows without `ok` and report `degraded: null`.

Description rewritten to match the final diff, including a new `degraded` value, one known limitation I did not fix (empty results carry no marker — return-shape decision, yours to make), and honest verification numbers.

One open question I would rather ask than guess: for the unenforceable cases I chose to leave the rows exactly as FTS5 answered and only mark them. Narrowing them would be easy but would silently turn `OR` into `AND`. If you would rather those queries error out instead of returning a marked best-effort result, say so and I will change it.

PR #80, opened 2026-08-19
fix(app): replace per-path watching with one recursive fs.watch per root

### LINKED ISSUE #79: App watcher: chokidar 4 per-path watching exhausts FSEvents/mach resources on large transcript trees (EMFILE flood, partial coverage)
## What happens

The default watcher in `app/src/main/indexer-service.ts` uses chokidar 4, which — having dropped fsevents — watches every path individually on macOS: one `O_EVTONLY` descriptor per file plus one FSEventStream per directory. On a real transcript tree (~18.7k files + ~4.3k directories under `~/.claude/projects`) the watcher floods `Obelisk watcher failed: EMFILE: too many open files, watch` (~700 lines / 30 s) and only ~15.2k of the 18.7k file watches actually establish, so new-file coverage rides on the surviving directory watches.

Ruling out the usual suspects, measured on macOS (node 26):

- **Not RLIMIT_NOFILE**: with the launchd agent's soft/hard limit raised to 49152 (verified in-process), the flood continues at ~15.2k fds.
- **Not kqueue/libuv per-watcher limits**: a bare-node probe holds 20,000 single-file `fs.watch` instances (20k fds) without error, in the same launchd environment and limits.
- **Directory watches are the resource hog**: per-directory `fs.watch` (FSEventStream each) degrades super-linearly — ~7 ms each at 1.2k streams, minutes for 6k — pointing at a per-process FSEvents/mach ceiling that libuv surfaces as `EMFILE`.

## Proposal (implemented, PR ready)

Replace the per-path model with **one `fs.watch(root, { recursive: true })` per configured root**: a single FSEvents stream on macOS (native on Windows, supported on Linux since Node 20), O(1) descriptors regardless of tree size, and events still fire for files inside directories created after the watcher started. Event consumption is unchanged — the add/change/unlink trio already funneled into one `.jsonl`/`.json`-filtered callback, which `fs.watch`'s `(eventType, filename)` covers; write stability is already owned by the service's debounce + stability timers.

Measured after the swap on the same tree: **0 EMFILE lines, 43 fds total** (was ~15.2k), watch-triggered builds fire as before.

chokidar would remain in use for the shallow `OBELISK_DIR` watch in `app/src/main/index.ts` (out of scope).

Cyb3rN8:
Fixes #79.

chokidar 4 (no fsevents) watches every path individually: one `O_EVTONLY` fd per file, one FSEventStream per directory. On a ~23k-path transcript tree the watcher floods `EMFILE: too many open files, watch` and leaves ~20% of files unwatched. Reproduction data in #79: rlimit ruled out (49152 verified in-process), bare `fs.watch` × 20k files ruled out (no error), per-directory stream creation degrades super-linearly (~7 ms each at 1.2k, minutes at 6k) — a per-process FSEvents/mach ceiling that libuv surfaces as EMFILE.

## Change

One `fs.watch(root, { recursive: true })` per configured root (files get a plain watch). Single FSEvents stream on macOS, native recursive on Windows, supported on Linux since Node 20. O(1) descriptors; events fire for files in directories created after the watcher started. Event semantics unchanged: all event types funnel into the same `.jsonl`/`.json`-filtered `onChange` — exactly how the add/change/unlink trio was consumed before. Write stability stays with the service's debounce + stability timers (chokidar's `awaitWriteFinish` was a second, redundant layer). A synchronous `fs.watch` failure logs and returns false, so the existing `refreshMissingRoots` retry loop owns recovery.

chokidar remains a dependency: `app/src/main/index.ts` still uses it for the shallow `OBELISK_DIR` watch, deliberately out of scope (one PR, one thing).

## Existing-assertion changes (own section per the contributing guide)

The four tests that injected a `chokidar` mock now inject `watchFs` (same shape as `fs.watch`, the new narrow injection point). Assertions on chokidar-specific options (`cwd`/`ignoreInitial`/`awaitWriteFinish`) became assertions on `{ recursive: true }`; the behaviors those tests pin — multi-root fan-out, late-appearing-root pickup, relative-path resolution, close-on-stop — are asserted unchanged.

## Verification

- New `tests/app-indexer-watch-recursive.test.mjs` drives the **real** watcher over a populated temp tree: deep new file in a directory created after watch start reaches `buildIndex` with an absolute path; non-transcript writes are filtered; watching a 100-entry tree costs ≤5 fds (the O(1) property that motivated the change).
- Full suite on this branch head: **467/467 pass, typecheck clean, lint 0 errors** (4 pre-existing warnings in unrelated tests, present on main).
- Not run: `test:electron:all` — this machine runs the data layer headless and has no Electron environment; the change is main-process-only (no renderer surface). Flagging per the verification contract so CI / a maintainer run covers it.
- Real-machine effect (headless daemon, same tree): EMFILE lines 697/30s → **0**; process fds ~15.2k → **43**; watch-triggered incremental builds fire as before.

### DIFF
--- a/app/src/main/indexer-service.ts
+++ b/app/src/main/indexer-service.ts
@@ -4,7 +4,6 @@
 import fs from 'node:fs';
 import os from 'node:os';
 import path from 'node:path';
-import chokidarModule from 'chokidar';
 
 const DEFAULT_PROJECTS_DIR = path.join(os.homedir(), '.claude', 'projects');
 const DEFAULT_DEBOUNCE_MS = 2000;
@@ -54,7 +53,8 @@ interface IndexerServiceOptions {
   buildIndex?: IndexerBuild;
   writeHeartbeat?: () => unknown;
   watchProjects?: (onChange: (changedPath?: string) => void) => Watcher | null;
-  chokidar?: any;
+  /** Injection point for the default watcher's fs.watch (tests). */
+  watchFs?: typeof fs.watch;
   timers?: Timers;
   logger?: { warn?: (msg: string) => void };
 }
@@ -70,7 +70,7 @@ function createIndexerService({
   buildIndex,
   writeHeartbeat = () => {},
   watchProjects,
-  chokidar,
+  watchFs = fs.watch,
   timers = {
     setTimeout,
     clearTimeout,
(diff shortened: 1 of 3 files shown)


tommy0103:
Thanks again for the investigation behind this — the diagnosis held up and directly informed the fix. We're closing this in favor of #83, which landed the same core idea (one recursive native watch per root, O(1) descriptors) but on top of [@parcel/watcher](https://github.com/parcel-bundler/watcher) instead of bare `fs.watch`, plus the two robustness layers discussed in #79 that this PR left open:

- **Dead-stream recovery**: a subscription that errors asynchronously is dropped from the watched set and re-attached through the existing watch-retry loop (here, an errored `fs.watch` root stayed in `watchedRoots` forever, and `refreshMissingRoots` only covered creation-time failures).
- **Reconciliation**: a periodic full-inventory build (5 min) bounds worst-case staleness from silently dropped events — FSEvents overflow, Windows buffer overflow, and the Linux `recursive` race window all lose events without any signal, so a watcher-only design can go stale permanently.

Using @parcel/watcher also means the Linux per-directory inotify emulation lives in its C++ layer (the same code VS Code runs at monorepo scale) rather than in Node core's JS fallback, and its N-API prebuilds load in Electron without a rebuild step.

The fd/resource win this PR demonstrated (43 fds vs ~15.2k, no EMFILE) is preserved — #83 keeps the same one-stream-per-root model on macOS.

PR #97, opened 2026-08-24
perf(app): make Activity usage stats index-backed

SaladDay:
## Summary

- Replace three repeated usage aggregates with one snapshot query, then derive lifetime and peak totals from the daily rows.
- Add partial covering indexes for token activity and longest-turn lookups.
- Preserve the existing treatment of undated token usage and accept raw provider timestamp strings.

## Scope

This PR only changes the `db:getUsageStats` query, its supporting indexes, and regression tests. It does not change the Activity renderer, provider indexing, background workers, caching, or other database queries.

## Schema decision

The schema change adds two non-destructive indexes. It does not change table layouts or rewrite indexed transcript rows. Existing databases build the indexes once through the current schema setup path; new and rebuilt databases create them with the rest of the schema.

On a local 2.4 GB database with 713,983 messages, the Activity query dropped from about 22.4 seconds to 0.14 seconds. The indexes use about 16.9 MB. Building them on that database took about 12 seconds once.

## Verification

- `npm test`: 409 passed
- `npm run typecheck`: passed
- `npm run lint`: 0 errors, 4 pre-existing warnings
- Production Electron build: passed
- `git diff --check`: passed
- New and old usage queries returned identical daily rows and lifetime totals on the benchmark database

`npm run test:electron:all` passed 4 of 5 local suites. `electron-session-virtualization.mjs` timed out while waiting for an existing SessionDetail timing assertion. The same timeout reproduces from a clean `origin/main` worktree, and this PR does not change renderer or SessionDetail code.

### DIFF
--- a/app/src/main/index.ts
+++ b/app/src/main/index.ts
@@ -771,60 +771,49 @@ ipcMain.handle('db:getStats', (_, opts = {}) => {
 
 ipcMain.handle('db:getUsageStats', (_, opts = {}) => {
   if (!db) return { daily: [], totalTokens: 0, peakDay: null, longestTurn: null };
-  const sourceFilter = sourceWhereClause(opts, 'source');
-  const sourceSql = sourceFilter.sql ? `AND ${sourceFilter.sql}` : '';
-  const usageEvents = `
-    WITH usage_events AS (
-      SELECT timestamp, input_tokens, output_tokens, COALESCE(source, 'claude') AS source
-      FROM messages
-      UNION ALL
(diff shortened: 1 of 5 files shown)


tommy0103:
Verified the `electron-session-virtualization.mjs` timeout no longer reproduces on this branch. The fixture fix from #106 (e8e826c) is included in the current head (b4d538c) via the main merge, and `npm run test:electron:timeline` now passes all 28 assertions locally, including the SessionDetail timing assertions that previously timed out. The "4 of 5 suites" note in the PR description is stale and can be dropped.

PR #110, opened 2026-08-25
fix(core): semantic read-only SQL validation for sql() (#107)

### LINKED ISSUE #107: [Query] Eliminate false positives in read-only SQL validation
## TL;DR

`sql()` rejects genuinely read-only queries when mutation keywords appear inside string literals, comments, or quoted identifiers. Validation should follow the statement's actual database effects, reject writes and ambiguous multi-statement input, and stay within an agreed sub-microsecond performance budget.

## Problem

CodeAct query scripts are authored dynamically by agents, so `sql()` needs a reliable fail-closed read-only contract. The current validator scans the complete SQL text for mutation keywords without distinguishing executable syntax from literals, comments, or identifiers. This creates avoidable retry loops for valid queries and makes the runtime contract depend on wording inside data-matching predicates rather than on what SQLite will execute.

## Current Behavior

`packages/core/src/query.ts:117-125` first requires a `SELECT`/`WITH` prefix, then rejects the entire input when a regular expression finds `INSERT`, `UPDATE`, `DELETE`, or another blocked token anywhere in the string. For example, this read-only query fails before SQLite prepares it:

```js
return sql("SELECT 'live update' AS text");
```

The query runtime already opens its database through `openReadDb()` with `{ readOnly: true }`, but the lexical pre-check prevents that database-level protection from seeing valid statements.

## Expected Behavior

- Read-only `SELECT`/`WITH` statements are accepted even when blocked words occur only in literals, comments, or quoted identifiers.
- Statements with write effects fail closed without mutating the index.
- A single `sql()` call has an unambiguous one-statement contract.
- The stronger semantic coverage does not add material query overhead.

### Agent-generated query shape

An Obelisk history scan excluded the invoking session and parsed actual `.mjs` files created by Codex, Claude, Kimi, and DeepSeek:

- 144 created `.mjs` query artifacts across 28 sessions;
- 299 literal `sql()` invocations;
- 68 scripts contained multiple independent `sql()` calls;
- 0 individual `sql()` arguments contained more than one top-level SQL statement.

This supports enforcing one statement per `sql()` call without breaking observed agent behavior. It does not prohibit a script from making several independent calls.

### Performance envelope

A Node.js v25.9.0 microbenchmark compared the current path with a semantics-aware prepare-time prototype over 12 alternating rounds of 8,000 executions for four representative query shapes. Median relative overhead ranged from about 2% to 10%, while absolute overhead remained about 0.06-0.66 microseconds per call. The accepted budget is:

- median added latency no greater than 1 microsecond per `sql()` call, measured as the workload median of per-shape median deltas; and
- relative regression no greater than 10% for the representative workload, measured as the workload-aggregate ratio (sum of per-shape median deltas over sum of per-shape legacy medians). Per-shape relative medians are not repeatable — ±100 ns of timer noise on a ~2 µs baseline amplifies into ±10% swings — so the aggregate ratio is the stable operationalization; see ADR 0010.

## Impacted Areas

| File | Lines | Current state |
|------|-------|---------------|
| `packages/core/src/query.ts` | 117-125, 153-156 | Lexically classifies SQL before preparing and executing it. |
| `packages/core/src/db.ts` | 38-43 | Opens the query database read-only; this remains the final mutation safety net. |
| `packages/core/src/sqlite-types.ts` | 10-24 | The shared statement/database shape currently exposes execution methods but no read-only classification metadata. |
| `tests/query.test.mjs` | Query API coverage | Needs regression coverage for lexical false positives, writes, and multi-statement input. |

## Reproduction

1. Run an Obelisk query script containing `return sql("SELECT 'live update' AS text")`.
2. Observe `sql() only supports read-only SELECT/WITH queries`.
3. Change the literal to text without a blocked keyword and observe that the same query shape succeeds.

## Acceptance Criteria

- [ ] Read-only queries containing `update`, `delete`, or other blocked words only in a string literal, SQL comment, or quoted identifier succeed.
- [ ] Top-level writes, including write statements prefixed by `WITH`, fail without changing the database.
- [ ] Input containing more than one SQL statement fails clearly; multiple separate `sql()` calls in one `.mjs` script remain supported.
- [ ] The query database continues to be opened read-only as the final mutation safety boundary.
- [ ] A checked-in, repeatable benchmark documents the comparison methodology and stays within both agreed limits as operationalized above: workload-median added latency no greater than 1 microsecond per `sql()` call and workload-aggregate relative regression no greater than 10% (see ADR 0010 for the measurement rationale).
- [ ] `npm test` passes.
- [ ] `npm run typecheck` passes.

## Related

- #96 tracks actionable CodeAct query diagnostics and does not overlap this issue's scope.

tommy0103:
Closes #107

## Summary

`sql()` no longer scans SQL text for mutation keywords (which false-rejected read-only queries like `SELECT 'live update' AS text`). Validation now follows the statement's actual database effects:

- **Top-level `SELECT`/`WITH` prefix contract kept.** It is the cheap, stable entry contract, and it keeps statement-level `PRAGMA` out of the sandbox while pragma table-valued functions (`pragma_table_info(...)`) inside a `SELECT` keep working.
- **Denylist authorizer (node:sqlite).** `createQueryApi` installs a prepare-time `setAuthorizer` write denylist: DML, DDL, `ATTACH`/`DETACH`, and `SAVEPOINT` return `SQLITE_DENY`; everything else — recursive CTEs (`SQLITE_RECURSIVE`), FTS shadow-table reads, pragma TVFs (`SQLITE_PRAGMA`), unknown future read actions — is allowed by default, so read syntax cannot false-positive the way the keyword scan did. Only `SQLITE_DENY` is ever returned: `SQLITE_IGNORE` on `SQLITE_READ` would silently null out columns instead of failing (verified). The authorizer is the semantic classifier; the `readOnly: true` connection from `openReadDb()` remains the final mutation boundary.
- **better-sqlite3 capability branch.** It exposes no authorizer, so classification there uses the statement's `readonly` flag; its native multi-statement rejection is preserved. The shared `SqliteDb`/`SqliteStatement` types gained optional capability fields (`setAuthorizer`, `readonly`, `sourceSQL`).
- **Multi-statement input fails clearly via `sourceSQL`.** node:sqlite's `prepare` compiles only the first statement and silently ignores the tail (verified: `SELECT 1; DROP TABLE t` runs only the `SELECT`), so the authorizer alone cannot see a second statement. `stmt.sourceSQL` exposes exactly the compiled portion; any tail beyond whitespace/comments throws `sql() accepts exactly one SQL statement per call; ...`. Multiple independent `sql()` calls per script remain supported, matching observed agent behavior (299 calls, 0 single-call multi-statements).

## Tests

New regression coverage in `tests/query.test.mjs`:

- blocked keywords in literals / line & block comments / quoted identifiers succeed (the issue repro)
- blocked word inside a `LIKE` literal executes (returns no rows instead of being rejected)
- recursive CTE and `pragma_table_info` TVF succeed
- `WITH ... INSERT`, `WITH ... DELETE` fail without mutating the database (asserted on a writable in-memory fixture where the authorizer exists, i.e. Node 24.10+, so the semantic checks — not the read-only connection — are what blocked them)
- a file-backed read-only fixture proves the final boundary on *every* supported Node version: a `WITH ... INSERT` fails (contract error at prepare on 24.10+, SQLite's raw read-only error at execute below it) and the data is unchanged
- `SELECT 1; DROP TABLE memories`, `SELECT 1; SELECT 2`, unterminated trailing comment fail clearly; `SELECT 1;` with trailing semicolon/comments succeeds

Latest head: `npm test` **504/504** pass (Node 25.9). `npm run typecheck`: pass. Changed files lint clean; full-repo `npm run lint` on tracked files reports 60 problems (0 errors, 60 warnings), all pre-existing and untouched by this PR (an earlier draft of this paragraph said 233 — that counted untracked local scratch directories). CI (Node 22 × ubuntu/macos/windows): green, including the new dedicated `tests/query.test.mjs` step.

## Benchmark

Checked in as `scripts/bench-sql-readonly.mjs` (repeatable, exits 1 when over budget). Legacy path is a frozen inline copy of the pre-#107 validator; current path is the real `sql()` from `createQueryApi`; 12 rounds per shape against identical in-memory DBs, block order flipping every round so machine drift cancels. Three isolated runs at head:

| shape | legacy ns/call | current ns/call | delta ns | delta % |
|---|---|---|---|---|
| trivial | 1986–2018 | 2102–2142 | ~120 | ~6.0 |
| aggregate | 2536–2627 | 2897–2984 | ~360 | ~13.9 |
| fts-join | 10058–10113 | 11075–11092 | ~1000 | ~9.9 |
| cte | 11776–11820 | 12360–12393 | ~580 | ~4.9 |
| sessions-list (1k rows) | 97715 | 97985–98505 | ~890 | ~0.9 |

**Gate results across the three runs: absolute +548/+681/+598 ns per call (workload median; budget ≤1 µs), relative +2.7/+2.8/+2.3% (workload-aggregate ratio; budget ≤10%). PASS, PASS, PASS.**

Methodology notes, documented in the script header:

- Overhead is a fixed ~90 ns per authorizer action crossing; a prepare fires 1–12 crossings depending on tables/columns/FTS shadow tables touched. Light shapes therefore show the highest *relative* numbers (aggregate ~14% at only ~0.36 µs absolute) while heavy realistic queries sit at ~1%.
- The FTS join is the boundary shape on both axes (~1.0 µs, ~10%) because FTS5 shadow tables add crossings; this mirrors the prototype numbers in the issue discussion (top of its reported 2–10% range).
- Gates: absolute = workload median of per-shape median deltas; relative = workload-aggregate ratio (Σ deltas / Σ legacy baselines). A per-shape relative gate is not repeatable — on ~2 µs-baseline shapes ±100 ns of timer noise amplifies into ±10% swings and identical code flickers across the 10% line (observed: 17.7% / 7.3% / 7.9% on the same commit). The aggregate ratio averages that noise over the full workload's runtime and matches the budget's intent: does the representative workload get materially slower. Per-shape rows remain printed for transparency. At real index scale (queries of tens of µs and up), measured relative impact is ≲1%.

### DIFF
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -39,6 +39,9 @@ jobs:
           npm run build --workspace @obelisk-apps/adaptive-watcher
           node --experimental-strip-types --experimental-sqlite --experimental-test-module-mocks --test tests/adaptive-watcher.test.mjs tests/core-watch-hints.test.mjs tests/claude-parse.test.mjs tests/app-indexer-watcher.test.mjs tests/app-indexer-service.test.mjs tests/app-main-settings.test.mjs
 
+      - name: Verify query sandbox read-only contract
+        run: node --experimental-strip-types --experimental-sqlite --experimental-test-module-mocks --test tests/query.test.mjs
+
       - name: Verify POSIX bootstrap installer
         if: runner.os != 'Windows'
         run: node --test tests/cli-bootstrap-install.test.mjs
--- a/docs/adr/0010-semantic-read-only-sql-validation.md
+++ b/docs/adr/0010-semantic-read-only-sql-validation.md
@@ -1,0 +1,98 @@
+# Semantic read-only validation for the query sandbox
+
+**Context.** Issue #107: `sql()` classified statements with a lexical
+pre-check — a `SELECT`/`WITH` prefix regex plus a mutation-keyword scan over
+the entire SQL text. The keyword scan cannot distinguish executable syntax
+from data, so read-only queries were rejected whenever a blocked word appeared
+inside a string literal, comment, or quoted identifier
+(`SELECT 'live update' AS text`, `substr(replace(text, ...))`, a `CASE WHEN`
+label containing the standalone word "replace"). The runtime contract depended
+on the wording of data-matching predicates instead of on what SQLite would
+execute.
+
+Three structural facts shaped the fix:
+
+- The query runtime always opens the index through `openReadDb()` with
+  `{ readOnly: true }`, so the lexical scan never was the mutation boundary —
+  it only produced early errors, and it produced wrong ones.
+- History shows a single-statement contract is safe: 299 historical `sql()`
+  calls contained zero multi-statement single calls, and a replay of all 435
+  unique literal SQL strings agents have written on the maintainer's machine
+  produced zero regressions on organic queries (PR #110).
+- The supported runtime floor (Node 22.13.0) does not offer a uniform
+  semantic classifier. `sqlite.constants` exists since v22.13.0 and
+  `StatementSync.sourceSQL` since v22.5.0, but `DatabaseSync.setAuthorizer`
+  exists only since v24.10.0; better-sqlite3 has no authorizer at all but
+  exposes `statement.readonly` and rejects multi-statement prepares natively.
+  Two traps were verified empirically: node:sqlite's `prepare` compiles only
+  the first statement and silently ignores the tail (so no prepare-time
+  mechanism can see a second statement), and returning `SQLITE_IGNORE` for a
+  `SQLITE_READ` action silently nulls columns instead of failing.
+
+**Decision.** `sql()`'s read-only contract is enforced in layers, each with a
+distinct job (`packages/core/src/query.ts`):
+
+1. **Top-level `SELECT`/`WITH` prefix check (all runtimes).** The cheap entry
+   contract. It also keeps statement-level `PRAGMA` out, which matters because
+   layer 2 must allow `SQLITE_PRAGMA` actions for pragma table-valued
+   functions (`pragma_table_info(...)`) used for schema exploration.
+2. **Prepare-time denylist authorizer where the driver exposes it**
+   (node:sqlite ≥24.10). DML, DDL, `ATTACH`/`DETACH`, and `SAVEPOINT` action
+   codes return `SQLITE_DENY`; everything else — recursive CTEs
+   (`SQLITE_RECURSIVE`), FTS shadow-table reads, pragma TVFs, and unknown
+   future read actions — is allowed by default. A denylist can never
+   false-positive on read syntax the way the keyword scan did, and failing
+   open on unknown actions is safe because of layer 4. Only `SQLITE_DENY` is
+   ever returned — never `SQLITE_IGNORE`. Authorizer denials surface as the
+   stable contract error (`sql() only supports read-only SELECT/WITH
+   queries`).
+3. **Multi-statement detection via `sourceSQL`** (node:sqlite ≥22.5).
+   `stmt.sourceSQL` exposes exactly the compiled first statement; any tail
+   beyond whitespace and comments fails with a distinct clear error. Multiple
+   independent `sql()` calls per script remain the supported way to run
+   several statements.
+4. **The read-only connection is the final mutation boundary (all
+   runtimes).** Anything layer 2 misses fails at execute time with SQLite's
+   own read-only error and the index never mutates. On Node 22.13–24.9 this
+   layer is also the write classifier: error-message quality degrades (raw
+   `attempt to write a readonly database` instead of the contract message) but
+   no write can succeed. The boundary never fails open.
+
+The capability seam in `packages/core/src/sqlite-types.ts` — optional
+`SqliteDb.setAuthorizer`, `SqliteStatement.readonly`, and
+`SqliteStatement.sourceSQL` — is deliberately reserved. The better-sqlite3
+`readonly` branch has no production consumer today (the app never runs
+`createQueryApi`), and review flagged it as possible speculative generality.
+It stays by maintainer decision: it is three lines completing an
+already-shared interface, and it is the only driver-correct classification
+for that binding if the app ever executes sandbox queries. This paragraph
+records that the reservation is intentional, not an oversight.
+
+This refines ADR 0002's Tier-1 contract. What is frozen is the sandbox
+guarantee — `sql()` is read-only and accepts exactly one statement per call —
+not the lexical mechanism that used to approximate it. The observable behavior
+change (previously rejected read-only queries now succeed) is the purpose of
+issue #107.
+
+The performance budget from issue #107 — no more than 1 µs median added
+latency per `sql()` call and no more than 10% relative regression on the
+representative workload — is gated in `scripts/bench-sql-readonly.mjs` as:
+absolute, the workload median of per-shape median deltas; relative, the
+workload-aggregate ratio (sum of per-shape median deltas over sum of legacy
+medians). Per-shape relative medians were measured to be non-repeatable
+(±100 ns of timer noise on a ~2 µs baseline amplifies into ±10% swings, and
+identical code was observed at 17.7% / 7.3% / 7.9%), so the issue's "median
+relative regression" is operationalized as the aggregate ratio, which averages
+that noise over the full workload and directly answers whether the workload
+gets materially slower. Per-shape numbers remain printed for transparency.
+
+**Consequences.** Keyword false positives are eliminated structurally, and
+read syntax added by future SQLite versions cannot be rejected by omission.
+The cost is a fixed ~90 ns per authorizer action crossing, 1–12 crossings per
+prepare depending on tables, columns, and FTS shadow tables touched; measured
+workload medians are ~0.5–0.7 µs absolute and ~2–3% aggregate-relative, with
+FTS-heavy shapes at the budget boundary (~1 µs). Tests are capability-gated:
+prepare-time write-rejection assertions require Node ≥24.10, while a
+file-backed read-only test proves the final boundary on every supported
+runtime. Multi-statement detection depends on node:sqlite's `sourceSQL`
+semantics, which the tests pin explicitly.
--- a/packages/core/src/query.ts
+++ b/packages/core/src/query.ts
@@ -169,9 +169,12 @@ const DENIED_SQLITE_ACTIONS: ReadonlySet<number> = new Set([
   sqliteConstants.SQLITE_SAVEPOINT,
 ]);
 
-// Prepare-time semantic classification. node:sqlite exposes the authorizer;
-// better-sqlite3 does not, and there statement classification happens through
-// the statement's readonly flag in assertReadOnlyStatement below.
+// Prepare-time semantic classification. node:sqlite exposes the authorizer
+// (setAuthorizer, Node 24.10+); on older supported runtimes and on drivers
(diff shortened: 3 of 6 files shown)


tommy0103:
## Real-corpus regression probe (reference data, not part of the test suite)

Replayed every literal SQL string agents have historically passed to `sql()` on this machine through both the legacy lexical validator and this PR's semantic validator, against a throwaway copy of the live index DB. Script: `tmp/sql-regression/run.mjs` (kept out of the tree on purpose).

### Corpus

- 349 tool calls (Write/Edit/Bash/exec across Codex/Claude/Kimi sessions) → **435 unique literal SQL strings** after decoding diff encodings and nested shell escaping.

### Verdict comparison (legacy vs new)

| bucket | count | notes |
|---|---|---|
| both accept | 237 | unchanged |
| legacy rejected, new accepts — **false positives fixed** | **10** | incl. the issue repro `SELECT 'live update' AS text`, a production query using `substr(replace(text, ...))` (function *name* `replace` tripped the old keyword regex), and two `CASE WHEN ... LIKE '%String to replace not found%'` classifiers (standalone word "replace" inside a literal) |
| legacy accepted, new rejects | **3** | all three are this PR's own multi-statement test fixtures (`SELECT 1; SELECT 2`, `SELECT 1;; SELECT 2`, `SELECT 1; /* unterminated`) that got indexed into the corpus; the legacy validator *accepted* `SELECT 1; SELECT 2` and silently ran only the first statement — rejecting these is the intended #107 behavior change. **Organic agent queries: zero regressions.** |
| both reject | 72 | one genuine `INSERT` (old test fixture), the new `WITH ... INSERT/DELETE` test fixtures, and corpus extraction residue (mangled diff text) — not real queries |
| both pass validation, exec error on current schema | 113 | schema drift (historical queries referencing since-changed columns); unrelated to validation |

### Performance on a copy of the live ~1 GB index (105 executable queries)

- legacy p50 = 43.8 µs/call; **added-cost p50 = +625 ns (+2.06 %)**
- p95 = +771 µs / +14.8 %, max outliers in the hundreds of ms — that spread is timer/GC/cache noise on multi-ms queries (an old-vs-old control run showed the same ±0.5 µs noise floor at 20 µs scale), not validation cost
- 48/105 queries added >1 µs, consistent with the ~90 ns × authorizer-crossings model (FTS-heavy shapes fire ~12 crossings per prepare)

Same conclusion as the checked-in microbenchmark: fixed per-prepare cost, median real-world impact ~0.6 µs / 2 %, invisible on heavy queries.

tommy0103:
## Response to review

**Spec P1 (writes on Node >=22.13.0) — the claimed security hole does not exist.** Verified facts:

- `setAuthorizer` was added in Node 24.10.0, `sqlite.constants` in v22.13.0/v23.5.0, and `StatementSync.sourceSQL` in v22.5.0 ([Node.js sqlite docs](https://nodejs.org/api/sqlite.html)). So on the supported floor (22.13.0) the module loads, the prefix contract holds, and the multi-statement tail check works.
- In production, `sql()` always runs against `openReadDb()`, which opens the index with `{ readOnly: true }` (`packages/core/src/db.ts:40`). Empirically, on a read-only connection **without** any authorizer, `WITH c AS (SELECT 1) INSERT INTO t SELECT * FROM c` throws `attempt to write a readonly database` at execute time and the table is unchanged. Writes fail closed on every supported Node version — the authorizer is the early, clearly-worded classifier, not the mutation boundary. "WITH ... INSERT executes successfully" cannot happen through any production path.
- What *does* degrade on Node 22.13–24.9: the error is SQLite's raw read-only error instead of the contract message, and the keyword false-positive fix needs 24.10+ to reject writes at prepare time. That is a message-quality degradation, not a safety hole. This is now documented at `installWriteDenylist`.

That said, the review exposed three real defects, fixed in 1dd94dd:

1. **Version-fragile tests** (the actual cause of the local 501/503): the write-rejection test asserted authorizer behavior on a *writable* in-memory fixture, which only holds where `setAuthorizer` exists. Those assertions are now capability-gated, and a new file-backed test (`a read-only connection fails writes closed on any supported Node version`) proves the final boundary on all supported runtimes — it accepts either the contract error (24.10+) or SQLite's raw read-only error (<24.10).
2. **CI coverage**: `tests/query.test.mjs` was missing from the workflow's explicit test list. Added a dedicated "Verify query sandbox read-only contract" step to `.github/workflows/cli.yml` (Node 22 matrix, all OSes).
3. **Benchmark fairness**: the header claimed alternating execution but the loop always ran legacy first. Block order now flips every round so systematic drift cancels; re-run after the change: workload median **+521 ns/call, +6.6%** — still within budget.

**macOS CI failure** is `tests/adaptive-watcher.test.mjs` ("default watcher filters non-transcript files", FSEvents timing) — untouched by this diff, main's recent runs are green, and ubuntu/windows passed on the same commit. Re-triggered to confirm flakiness.

**better-sqlite3 `readonly` branch (judgment item)**: kept deliberately — it is three lines implementing the other half of the already-shared `SqliteDb` interface, and removing it would leave the type contract documenting a capability nothing implements.

Local verification on Node 25.9: `npm test` 504/504, `npm run typecheck` pass, changed files lint clean, benchmark PASS.

tommy0103:
## Response to review round 2

**Spec P1 (benchmark repeatability) — agreed in substance, fixed by changing the gate metric rather than the implementation.** The overhead is a fixed ~90 ns per authorizer action crossing; the absolute gate never moved (+461–681 ns, budget 1 µs). What flickered was the per-shape *relative* median: on shapes with a ~2 µs baseline, ±100 ns of timer noise amplifies into ±10% swings, and the workload median of five shape medians straddled the 10% line. The relative gate now uses the workload-aggregate ratio (Σ per-shape median deltas / Σ legacy medians), which averages the noise over the full ~124 µs workload and directly answers the budget's intent — whether the representative workload gets materially slower. Methodology and rationale are documented in the script header. Three isolated runs at head (3ed7c66): **+548/+681/+598 ns, +2.7/+2.8/+2.3% — stable PASS**, no flicker.

**Standards P2 (stale verification record) — fixed.** PR body updated: latest head is `npm test` 504/504, typecheck pass, CI green on all three Node 22 platforms including the new `tests/query.test.mjs` step. Full-repo `npm run lint` status is now stated explicitly: 233 pre-existing problems in unrelated/untracked areas; this PR's changed files are lint clean and fixing the pre-existing set is out of scope here.

**better-sqlite3 `readonly` branch** — unchanged from last round's rationale: it completes the already-shared `SqliteDb` interface in three lines; happy to drop it if the maintainer prefers, but it is the only driver-correct classification for that binding.

tommy0103:
## ADR added + review round 3 items

- **`docs/adr/0010-semantic-read-only-sql-validation.md`** (6d7a430) records the decision set: the layered enforcement and which layer owns which guarantee; the verified traps (node:sqlite `prepare` silently ignores statement tails, `SQLITE_IGNORE` on `SQLITE_READ` nulls columns); the per-runtime degraded modes (Node 22.13–24.9 blocks writes at execute via the read-only connection — degraded message, never fail-open); the **deliberately reserved better-sqlite3 `readonly` capability seam** — kept by maintainer decision, documented as intentional rather than an oversight; and the operationalization of #107's "median relative regression" as the workload-aggregate ratio, with the measurement evidence for why per-shape relative medians are not repeatable. It also notes this refines ADR 0002's Tier-1 contract: frozen is the guarantee (read-only, one statement per call), not the lexical mechanism.
- **Standards (full-lint number) — corrected in the PR body.** Tracked files only: **60 problems, 0 errors, 60 warnings**, all pre-existing. The earlier 233 figure counted untracked local scratch directories (`tmp/`, `website/`, draft docs) that do not exist in a clean checkout; that was my measurement error.

PR #116, opened 2026-08-28
fix(app): close source filter on outside click

cunninghamcard-bit:
## Reproduction

Open the source filter, then click elsewhere in the window. The popover remains open.

## Root cause

The popover visibility was only changed by its trigger and menu choices; it had no outside-boundary listener.

## Fix

Track the source-filter wrapper and close the popover on a document `pointerdown` whose target is outside that wrapper. The listener is added and removed with the component lifecycle.

## Verification

- `npm test`: 546/546 passed.
- `npm run typecheck`: root and app tsconfigs passed (2/2).
- `npm run lint`: 0 errors, 9 existing warnings.
- `npm run test:electron:all`: 4/5 suites passed locally; `electron-session-virtualization.mjs` failed macOS hidden-window wheel/timing assertions. The unchanged `origin/main` baseline also produced 4/5 in the same environment (virtualization failed with `patchReads: 0`). No test assertion was changed or loosened.

## Scope

This PR now contains only the outside-click fix. Breadcrumb truncation was split into #117.

### DIFF
--- a/app/src/renderer/src/App.vue
+++ b/app/src/renderer/src/App.vue
@@ -199,10 +199,12 @@ function handleGlobalKeydown(event) {
 
 onMounted(() => {
   window.addEventListener('keydown', handleGlobalKeydown);
+  document.addEventListener('pointerdown', handleDocumentPointerDown);
   stopSourceUpdates = window.obelisk?.onIndexUpdated?.(() => loadSourceDots()) ?? null;
 });
 onUnmounted(() => {
   window.removeEventListener('keydown', handleGlobalKeydown);
+  document.removeEventListener('pointerdown', handleDocumentPointerDown);
   stopSourceUpdates?.();
   stopSourceUpdates = null;
   clearTimeout(searchTimer);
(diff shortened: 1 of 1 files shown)


cunninghamcard-bit:
One small design question: this PR currently truncates long project names with a trailing ellipsis, while keeping the full name in a tooltip. The ellipsis may not be the best-looking treatment here—would you prefer middle truncation, a shorter fixed-width label, or another approach? Happy to adjust it to match your preferred visual style.

tommy0103 (commented):
Thanks for the fix — the outside-click close is solid (pointerdown-before-click ordering is handled correctly, and the listener add/remove is symmetric).

One request: could you split this into two PRs? Per CONTRIBUTING.md ("One PR does one thing"), the popover fix and the breadcrumb truncation are two unrelated user-visible changes. The popover half looks ready to go on its own.

For the breadcrumb half, there is a side effect to fix before it lands: adding `overflow: hidden` to the shared `.breadcrumb` container (toolbar.css) clips *every* breadcrumb branch, not just the project crumb. On SessionDetail/MemoryDetail/RecapDetail the terminal crumb (full session title, etc.) has no truncation class, so a long title that previously overflowed but stayed readable is now hard-clipped mid-glyph — no ellipsis, no tooltip. Either scope the clipping to the project-crumb branch, or give the title crumbs the same ellipsis + `:title` treatment.

Also note for the follow-up PRs: since this touches `app/`, the verification contract requires `npm test` and `npm run test:electron:all` (all five suites) with the numbers quoted in the description — "typecheck, lint, package build, manual UI verification" does not cover it.

cunninghamcard-bit:
Split as requested: this PR now contains only the source-filter outside-click fix. The project breadcrumb change is in #117, with clipping scoped to `.project-crumb` so the shared detail breadcrumbs remain unchanged. I also reran the required verification and quoted the results in both PR descriptions; the local macOS Electron result is 4/5 on both branches and on unchanged `origin/main`, with only the virtualization timing/wheel suite failing.

tommy0103 (commented):
Verification cross-check from a maintainer machine (macOS, Electron 43.2.0): I built this head (accd0bb) and ran `electron-session-virtualization.mjs` 4 times. Result: 3/4 runs fully green (55/55 assertions); one run failed a single perf-budget assertion — `stationary live commit 2 stays inside a 120Hz renderer task budget (42.24ms across 16 tasks)` — which did not reproduce on the other three runs and is a renderer-task timing measurement, not something this diff can affect (toolbar-only change; the suite mocks all IPC and never mounts the toolbar path this PR touches).

I could not reproduce the `patchReads: 0` baseline failure you saw: on the same machine, 5/5 runs of the suite pass on code identical to `origin/main` in renderer/preload. So that failure looks environment/timing-specific rather than a stable main-branch defect. Either way, nothing here suggests this PR regresses the suite — LGTM on the verification front.

Issue #114, opened 2026-08-26
[Codex] Guardian transcripts are never skipped in incremental discovery — hundreds of unchanged files re-parsed on every build


tommy0103:
# [Codex] Guardian transcripts are never skipped in incremental discovery — hundreds of unchanged files re-parsed on every build

## TL;DR

`discoverAt` refuses to skip guardian-thread transcripts even when their cursor and mtime are both clean, so every incremental build fully re-reads and re-parses every guardian file. On a real corpus (1,430 codex rollouts, 405 guardian threads = 28%) this costs ~28 s of a ~33 s otherwise-no-op build, and is the dominant per-query latency for CLI/plugin users who run without the app daemon.

## 问题描述

`packages/core/src/providers/codex.ts:100-103` skips an unchanged unit only when all of these hold:

```js
if (!sessionIndexChanged && !fileChanged && cursor !== null
    && Number(cursor.split(':')[0]) >= statSync(file.path).mtimeMs
    && guardian === null) {   // guardian files never skip
  return [];
}
```

For guardian transcripts (`codexIsGuardianThread`, `parsing.ts:265` — codex-auto-review subagents), `guardian !== null` permanently, so they are re-planned on every build. Each re-planned guardian file is then:

1. read in full during discover (`readCodexGuardianThreadInfo` scans records for `codex-auto-review`),
2. read and JSON.parsed in full again during `parse`,
3. persisted as an idempotent `delete-session` no-op.

Measured on a real-world index (~254k messages, 1,430 codex rollout files, 405 guardian):

| Component of one no-change incremental build | Wall time |
|---|---|
| guardian re-read + re-parse, 405 files (~74 ms/file) | ~28 s |
| `createProviderIndexPlan` (discovery walk) | ~2 s |
| unconditional `messages_fts` rebuild in finalize (#75) | ~2.2 s |
| everything else | < 0.1 s |

This also explains why per-build fixed-cost reports differ by an order of magnitude across machines: FTS-bound corpora see #75's 10–30 s, guardian-heavy corpora see this issue's ~28 s, and the two mask each other.

Every obelisk CLI query pays this when no app daemon owns the index (`refreshQueryIndex` builds synchronously before each query, and typical query spacing exceeds the 30 s build debounce), and the DSH plugin's default `timeoutMs: 30000` sits right at the edge of it.

## Root cause

[...]

Issue #104, opened 2026-08-24
[Codex] Detect same-mtime transcript changes in incremental discovery


tommy0103:
## TL;DR

Codex transcript discovery currently treats mtime alone as the persisted change signal. A transcript append or replacement that retains the same mtime can therefore remain absent from the index until some later detectable change occurs.

## 问题描述

The Codex provider performs a full-file parse once a unit is selected, so this does not carry Claude's permanent line-skip failure mode. The remaining problem is discovery: during reconciliation without an exact changed-file hint, a same-mtime update can be classified as unchanged and leave the indexed session temporarily stale. This conflicts with the repository's cursor contract for same-millisecond rewrites.

## 当前行为

`packages/core/src/providers/codex.ts:100-103` reads the first cursor field as mtime and skips discovery when it is not older than the file's current mtime. Other file identity/change metadata is not represented in that decision.

`CONTRIBUTING.md:137-138` requires cursors to detect same-millisecond rewrites and account for moves, copies, deletes, and replacements.

## 期望行为

Codex discovery should eventually select a transcript whose meaningful file state changed even when its mtime did not, so both watcher-hinted indexing and later full reconciliation converge on current content.

## 影响范围

| 文件 | 行号 | 当前状态 |
|------|------|---------|
| `packages/core/src/providers/codex.ts` | 100-103 | Reconciliation short-circuits using the cursor's mtime field alone |
| `CONTRIBUTING.md` | 137-138 | Defines the required cursor behavior for same-millisecond rewrites |

## 触发/复现方式

1. Index a Codex JSONL transcript and retain its persisted cursor.
2. Append to or replace the transcript while preserving its prior mtime.
3. Run discovery without an exact changed-file hint.
4. Observe that the transcript can be skipped and the indexed session remains stale.

## 验收标准

- [ ] A Codex transcript with changed content or file identity and unchanged mtime is rediscovered.
- [ ] Both exact changed-path indexing and full reconciliation converge on the updated transcript content.
- [ ] Append and replacement regressions demonstrate that the indexed session does not remain stale.
- [ ] Existing Codex discovery and parse behavior remains covered, including compatibility with persisted cursors.
- [ ] The new regression coverage runs in CI on the repository's supported platforms.
- [ ] `npm test`, `npm run typecheck`, and `npm run lint` pass.

## 关联

- Follow-up to #102

PR #122, opened 2026-08-30
fix(core): strengthen Codex cursors to the mtime+ctime+size+inode signature

tommy0103:
> Stacked on #121 — base retargets to main automatically once #121 merges.

## Background

#104: codex incremental discovery skips files on an mtime-only cursor comparison, so a same-millisecond rewrite (rapid append landing in the same mtime tick, or a same-size-at-same-mtime replacement) is silently missed. CONTRIBUTING requires cursors to detect same-millisecond rewrites via mtime + ctime + size + inode; claude already implements this, codex is the outlier.

This was originally part of #121 and was split out to keep that PR minimal. #121's marker bump to `__codex_canonical_transcript_v3__` is the migration carrier for this cursor-format change, as #114 anticipated.

## Changes

- `packages/core/src/parsing.ts`: `cursorSignatureDiffers` moved here from `claude.ts` so both providers share one implementation of the documented signature check. No behavior change for claude.
- `packages/core/src/providers/codex.ts`: the discovery skip check uses `cursorSignatureDiffers`; `parse()` returns `${mtime}:${lines}:${size}:${ctimeMs}:${ino}`. Legacy two-part cursors keep the mtime-only gate and upgrade on the next parse; the v3 marker replay rewrites them outright wherever codex sessions exist.
- `.github/workflows/cli.yml`: add `codex-parse.test.mjs` to the CI test list (it asserts the codex parse contract and was not covered).
- `tests/codex-discover.test.mjs`: cursor fixtures move to the five-part format; new case pins that a same-mtime/different-size guardian rewrite is re-planned.

Runtime cost of the stronger check is zero: the skip path already called `statSync` for mtime; size/ctime/ino come from the same call. The only behavioral trade-off is that metadata-only changes (e.g. `chmod`) now cause one conservative re-parse — the safe direction.

## Scope

This closes the *cursor* half of #104. The watcher-side reconcile of moves/copies/deletes mentioned in CONTRIBUTING remains #104 scope.

## Verification

- Affected suites: 45/45 pass (codex-discover, codex-parse, codex-index, claude-parse, app-indexer, runtime). Full suite on the identical code state (pre-split): 564/564.
- `npm run typecheck`: clean. `eslint` on changed files: clean.

Refs #104

### DIFF
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -42,7 +42,7 @@ jobs:
       - name: Verify adaptive watcher package
         run: |
           npm run build --workspace @obelisk-apps/adaptive-watcher
-          node --experimental-strip-types --experimental-sqlite --experimental-test-module-mocks --test tests/adaptive-watcher.test.mjs tests/core-watch-hints.test.mjs tests/claude-parse.test.mjs tests/codex-discover.test.mjs tests/app-indexer-watcher.test.mjs tests/app-indexer-service.test.mjs tests/app-main-settings.test.mjs tests/app-indexer-watcher-filter.test.mjs tests/deepseek-tree.test.mjs
+          node --experimental-strip-types --experimental-sqlite --experimental-test-module-mocks --test tests/adaptive-watcher.test.mjs tests/core-watch-hints.test.mjs tests/claude-parse.test.mjs tests/codex-parse.test.mjs tests/codex-discover.test.mjs tests/app-indexer-watcher.test.mjs tests/app-indexer-service.test.mjs tests/app-main-settings.test.mjs tests/app-indexer-watcher-filter.test.mjs tests/deepseek-tree.test.mjs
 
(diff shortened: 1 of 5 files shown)


tommy0103:
Folded back into #121: since #121 has not shipped, the cursor fix now rides the same v3 marker bump, and legacy codex cursors fail closed (one re-parse, then upgrade) so no v3→v4 sequencing is needed. The remaining review points from here were addressed there: fail-closed legacy cursors, real end-to-end same-mtime append/rewrite regressions, and CI coverage for the codex test files. Claude was kept untouched — its legacy gate stays as #102 designed it.

tommy0103:
Superseded by #123 (stacked on #121). The branch was recreated from the revised commit — codex-local predicate with fail-closed legacy cursors, claude untouched, real end-to-end same-mtime regressions, CI coverage for the codex test files.

PR #157, opened 2026-09-14
fix: guard parent-chain traversal against cycles

### LINKED ISSUE #156: [Query] Bound parent-chain traversal and stop cycles
## Problem

context() and trace() follow parent_uuid without cycle detection or a traversal limit.

A corrupt index can contain a self-cycle or a multi-message cycle. Both helpers then loop forever. A very deep chain also causes unbounded SQLite reads.

Issue #93 identified this risk while discussing the duplicate helpers. This issue keeps both public helpers and fixes their shared traversal behavior.

## Expected behavior

- Stop before returning a repeated UUID.
- Stop cleanly when a parent row is missing.
- Follow at most 1,000 parent edges.
- Keep root-first ordering.
- Keep the target in trace() only.
- Preserve visibility filtering.

## Verification

Add tests for a self-cycle, a two-message cycle, a missing parent, and a chain above the limit.

## Scope

Update the shared query implementation, tests, and query contract documentation. Do not change the database schema.

luojiyin1987:
## What and why

`context()` and `trace()` can loop indefinitely if a parent chain contains a cycle.
This change adds defensive protection. No production parent-cycle example has been reported in this PR.

Both helpers share a traversal that tracks visited UUIDs and stops before a repeated UUID.
Valid chains return complete results without a depth cap, including chains beyond 1,000 parent edges.
Results preserve root-first ordering and visibility filtering.
Missing parents continue to end traversal cleanly.

Both API references document the traversal behavior.

Closes #156

## Verification

- `node --test tests/query.test.mjs`: 35 passed, 0 failed.
- `git diff --check`: passed.
- The full test suite, typecheck, and lint were not rerun for this revision.

## Test coverage

- A target references itself.
- Two messages reference each other.
- A missing parent preserves existing termination behavior.
- A valid chain of 1,003 messages returns complete results through both public helpers.
- Existing tests verify visibility filtering.


tommy0103 (commented):
Thanks for the clean refactor and the thorough test matrix. I want to push back on the framing, though, based on what real obelisk data actually looks like.

I ran a full DFS over the parent graph of a production index (352,742 messages, all five sources — claude / codex / kimi / pi / deepseek):

- Self-parents: 0. Two-message cycles: 0. Cycles of any length: **0**.
- Missing parents: 5,616 rows — common in real data, but never a live bug: the old loop already terminated cleanly there (`get()` → `undefined` → `isQueryableMessage(undefined)` is false → loop ends).
- Deepest legitimate chain: **11,726 edges** (a long codex rollout). **41 sessions** have chains deeper than 1,000, covering 95,254 messages.

Structurally, cycles should not occur in obelisk data at all: `uuid` is a PRIMARY KEY written via upsert, and every provider derives `parent_uuid` either pass-through from the transcript (claude) or from sequential parse order (`lastMessageUuid` in codex) — parents always point to earlier-written messages. A cycle would require a corrupt or hand-edited session file.

Two consequences:

1. The infinite-loop scenario is a theoretical/defensive concern, not something observed or likely on real data. And the part of this PR that actually guards against it is the `visited` set — not the depth cap. A depth limit alone wouldn't fix a cycle anyway; it would just return 1,000 messages with repeats. Once the `visited` set exists, the cap adds no correctness.
2. The depth cap is the one change here that *does* affect real, correct usage: `trace()` / `context()` on those 41 deep sessions now silently return a truncated chain instead of root-to-target. It buys no cycle safety beyond what `visited` already provides, and it doesn't speed up the common case either (same number of reads up to the cap).

My suggestion: keep the `visited`-set cycle guard as cheap hardening, but drop `MAX_PARENT_EDGES` — or, if a backstop against pathological files feels necessary, set it well above observed real depths (e.g. 10k+) or surface an explicit "truncated" marker so callers can tell the chain is incomplete.

Also curious: did this come from an actually-observed corrupt index? If real data with a parent cycle exists in the wild, that changes the picture — I'd like to understand how it was produced.

luojiyin1987:
Thanks for checking this against production data. I have no observed parent-cycle case to report. This was defensive hardening based on a theoretical risk.
You’re right: the 1,000-edge cap can truncate valid chains. I chose that limit without production depth measurements.
I’ll remove the cap and keep the visited guard. I’ll also update the tests and documentation to require complete traversal of valid chains.
Missing-parent handling already worked. I’ll clarify that its test preserves existing behavior.

Issue #183, opened 2026-09-18
[Linux] Double-click on the titlebar no longer toggles maximize


tommy0103:
Follow-up from #178.

## Problem

With the frameless window on Linux, the window manager no longer provides double-click-to-maximize on the titlebar area. The in-app `.titlebar` is a drag region (`-webkit-app-region: drag`, `app/src/renderer/styles/base.css`), and double-clicking it does nothing — a behavior framed windows got for free.

## Expected

Double-clicking the in-app titlebar toggles maximize/restore, matching native titlebar behavior.

## Suggested fix

Add a `@dblclick` handler on `.titlebar` in `app/src/renderer/src/App.vue` that calls `windowControl('toggle-maximize')` through the `win:control` IPC proposed in https://github.com/tommy0103/obelisk/issues/176#issuecomment-5732734013. `.titlebar-text` is already `pointer-events: none`, so the event lands on `.titlebar` cleanly.

Gate to Linux (or non-macOS): macOS handles double-click on its hidden-inset titlebar natively.

Depends on the `win:control` IPC from the window-controls follow-up; natural to land in the same PR.

PR #186, opened 2026-09-19
fix(app): center the titlebar on platforms without traffic lights

### LINKED ISSUE #184: [Linux] Titlebar title sits off-center: 78px left padding is a macOS-only allowance
Follow-up from #178.

## Problem

`.titlebar` in `app/src/renderer/styles/base.css` has `padding: 0 16px 0 78px`. The 78px reserves space for the macOS traffic lights (`trafficLightPosition` in `createWindow()`). On Linux — and Windows — there are no traffic lights, so the `justify-content: center` title sits ~31px right of true center. Visible in the "after" screenshot of #178.

## Expected

Title is visually centered on platforms without traffic lights.

## Suggested fix

Platform-gate the padding: expose the platform via preload (`process.platform`) and add a modifier (e.g. `.titlebar.linux` with symmetric `padding: 0 16px`), keeping the 78px allowance only on macOS.

If the custom window controls from https://github.com/tommy0103/obelisk/issues/176#issuecomment-5732734013 land first, account for their width on the right (~3 × 40px) so the title stays centered between the controls and the left edge.

FATFATHAO:
## What and why

`.titlebar` in `app/src/renderer/styles/base.css` reserves 78px on the left for the macOS traffic lights. Linux and Windows have no traffic lights, so with `justify-content: center` the title sat right of true center — visible in the "after" screenshot of #178.

Stacked before/after of the same window's titlebar: on main the title sits 31px right of the window centre; with this PR it matches the centre

That image is two captures of the same window (the title is 157px wide in both), scanned along the titlebar band: the title centre moves from x=1310.5 to x=1279.5 against a window centre of x=1280, so the left/right gaps go from 1216/1154 to 1185/1185. The 62px difference between the gaps is the `78 − 16` from CSS, and the title moves by half of it — the 31px that #184 describes.

The platform now decides the inset: preload exposes `platform: process.platform`, `.titlebar` takes a `mac` modifier, and the 78px allowance moves into `.titlebar.mac`. Linux and Windows get symmetric `padding: 0 16px`; macOS keeps exactly the 78px it has today.

`platform` is the one new concept here. The renderer had no platform signal at all, and CSS alone cannot know whether traffic lights exist, so exposing the value suggested in #184 is the smallest change that keeps the allowance in one place instead of duplicating the number.

Closes #184

## Verification

- [x] `npm test` — 715 pass / 0 fail (Node v22.22.2, matching CI)
- [x] `npm run lint` — 0 errors, 9 pre-existing warnings
- [x] `npm run test:electron:all` — 2/5 suites, identical to the untouched `main` baseline in the same session
- [x] Ran the app on Linux and measured the captures above (`window.obelisk.platform === 'linux'`, computed `padding-left: 16px`)
- [x] Re-ran the checks above on the current head (`8b8dc39`); no merge occurred after verification

`npm run test:electron:all` cannot be fully green on Linux; untouched `main` (`aa06bd4`) produces the same numbers in the same niri/Wayland session:

| suite | `main` (aa06bd4) | this branch (8b8dc39) |
| --- | --- | --- |
| `electron-concurrency` | 15 pass / 0 fail | 15 / 0 |
| `electron-session-images` | 11 pass / 1 fail | 11 / 1 |
| `electron-session-virtualization` | 0 pass / 1 fail | 0 / 1 |
| `electron-session-reader-state` | 6 pass / 5 fail | 6 / 5 |
| `electron-file-references` | 30 pass / 3 fail | 30 / 3 |
| suites passed | 2/5 | 2/5 |

## Deliberately out of scope

- No `win:control` IPC, no buttons, no double-click handling. Those are #183 and land in a separate PR, so this one stays independently revertable.
- macOS is unchanged by value: the 78px allowance is the same number, moved into `.titlebar.mac`. I have no macOS machine, so that path is not exercised here — the diff only relocates the existing declaration.
- The title is centred on the window, not yet between a control cluster and the left edge. When the controls land, that PR reserves the ~3 × 40px on the right as #184 notes.

---

<details>
<summary><b>Renderer / Electron UI</b></summary>

- [x] Reader-anchor assertion — not applicable: `.titlebar` sits outside the virtualized list; no row rendering, sizing, or scroll compensation is touched
- [x] Async media at mount / size-available / load-error — not applicable, no media added
- [x] Tested at rest and mid-scroll — not applicable, the titlebar does not scroll; the captures were taken at rest
- [x] No `loading="lazy"` on virtualized rows — untouched
- [x] No hardcoded colors or type sizes — only the existing padding value is relocated; no new token
- [x] One visual treatment per user-visible concept — the titlebar is still the only title treatment
- [x] Library calling conventions — not applicable
- [x] Every renderer probe has a deadline and an `error` → reject path — not applicable, no new probe
- [x] No assertion with sub-pixel headroom — not applicable, no assertion added

</details>

<details>
<summary><b>Main process / untrusted input</b></summary>

- [x] Transcript-derived path handling untouched; `shell.openPath` call sites unchanged
- [x] No file-read path added
- [x] No `innerHTML` or text-decoding change
- [x] No synchronous IO added
- [x] Interception and navigation handling unchanged
- [x] `app/src/preload/index.ts` stays TypeScript (ADR-0005); the new field is a constant string and no untrusted input reaches it

</details>


tommy0103 (approved):
LGTM. Verified beyond CI:

- CSS specificity is correct: `.titlebar.mac` (0,2,0) overrides `.titlebar` (0,1,0), so macOS keeps the exact 78px declaration — relocated, not rewritten. Grep confirms nothing else in the repo depends on that padding value.
- No bridge type declaration exists to keep in sync (renderer is all JS; the preload literal typechecks on its own).
- The RecapExport route renders without `.titlebar`, so the offscreen export capture is unaffected.
- Ran the branch on macOS (`npm run dev`): traffic lights in place, title position identical to main — the `mac` path is exercised and correct.
- The before/after pixel measurement in the description is self-consistent (78−16=62, half is the 31px shift from #184).

Scope discipline appreciated: keeping `win:control`/buttons/double-click (#183) out of this PR keeps it independently revertable. When the controls PR lands, remember the ~120px right reservation noted in #184.

One note unrelated to this diff: during my macOS dev run the process died with a `@parcel/watcher` napi fatal error in the shutdown path (`CleanupHandles` → `getIgnoreGlobs`). The watcher teardown code is untouched here; I'll track that separately if it reproduces on main.

PR #188, opened 2026-09-19
fix(app): await watcher shutdown before quitting

### LINKED ISSUE #187: [macOS] Cmd+Q can crash: quit never awaits watcher close, @parcel/watcher callback fires into a tearing-down JS env
## Problem

Cmd+Q sometimes ends with the macOS "Electron quit unexpectedly" dialog. The main process aborts with:

```
FATAL ERROR: Error::ThrowAsJavaScriptException napi_throw
  ... Init(Napi::Env, Napi::Object) [@parcel/watcher-darwin-arm64/watcher.node]
  ... PromiseRunner::onWorkComplete
  ... callbackEventsToJS / callJSFunction
  ... Napi::ThreadSafeFunction::CallJS
  ... uv__async_io / uv_run
  ... node::Environment::CleanupHandles()
```

## Root cause

`packages/adaptive-watcher` is safe **only if `close()` is awaited**: `close()` sets `closed = true` synchronously (`packages/adaptive-watcher/src/index.ts:420`) and every parcel event callback guards on it (`index.ts:205`).

The app breaks that guarantee on the quit path (`app/src/main/index.ts:490`):

```ts
app.on('before-quit', () => {
  void stopBackgroundResources({ stopWorker: true });
});
```

`before-quit` has no async contract in Electron — the handler returns immediately and quit proceeds. `stopBackgroundResources` hits its first `await` (`service.idle()`) almost immediately, so `obeliskWatcher.close()` — the call that flips `closed` — has not run when Node starts `node::Environment::CleanupHandles()`.

During teardown libuv drains the still-live `@parcel/watcher` ThreadSafeFunction handles; a queued FSEvents batch fires the JS callback *while the environment is being destroyed*. `closed` is still false, so the callback proceeds into `onInvalidate` → `setTimeout(...)` and throws; a napi TSFN callback has no JS frame above it to receive the exception, so `napi_throw` escalates to `node::OnFatalError` → abort.

It is a pure timing race: quitting with FSEvents deliveries in flight (active transcripts, just after indexing) crashes; quitting fully idle usually exits clean. Same pattern exists on the non-macOS path: `window-all-closed` also fire-and-forgets the stop before `app.quit()` (`app/src/main/index.ts:494`).

Race has existed since the adaptive-watcher integration (ADR-0009); unrelated to #178 / #186.

## Suggested fix

Deferred quit: prevent the first `before-quit`, run the full stop, then re-quit.

```ts
let quitting = false;
app.on('before-quit', (event) => {
  if (quitting) return;
  quitting = true;
  event.preventDefault();
  void stopBackgroundResources({ stopWorker: true })
    .catch(() => {})
    .finally(() => app.quit());
});
```

No caching of the stop promise: a macOS pause (`window-all-closed`) can be followed by `activate` → `startBackgroundResources`, and a later quit must stop the *restarted* singletons. `stopBackgroundResources` is null-safe for no-op repeats.

## Verification

Timing-dependent native crash, so deterministic coverage is impractical in the hand-driven electron suites. Verification: typecheck/lint, plus manual — run the app, produce watcher activity (indexing / live transcript writes), Cmd+Q during it, confirm a clean exit (no crash dialog) repeatedly.

First observed in a dev run on PR #186's branch; reproduced by maintainer on Cmd+Q.

tommy0103:
## What and why

Closes #187.

Cmd+Q on macOS could end in the "Electron quit unexpectedly" dialog: the main process aborted with `FATAL ERROR: Error::ThrowAsJavaScriptException napi_throw` inside `@parcel/watcher`'s native module during `node::Environment::CleanupHandles()`.

Root cause chain:

- `packages/adaptive-watcher` is only safe if `close()` runs before teardown: it flips `closed` synchronously and every parcel event callback guards on it.
- `before-quit` fired and forgot the async stop (`void stopBackgroundResources(...)`), and Electron does not wait for async `before-quit` handlers. The function hit its first `await` (`service.idle()`) before reaching `obeliskWatcher.close()`, so quit proceeded with live FSEvents subscriptions.
- During env teardown, libuv drains the watcher's ThreadSafeFunction handles; a queued event batch fires the JS callback while the environment is being destroyed. `closed` is still false, the callback runs into `onInvalidate` → `setTimeout` and throws; a napi TSFN callback has no JS frame above it, so the exception escalates to `node::OnFatalError`.

Fix, in `app/src/main/index.ts`:

1. `before-quit` defers the first quit request via `createDeferredQuit` (`app/src/main/quit-teardown.ts`): `preventDefault()`, run the stop, re-quit once it settles. Bounded at 5s so a wedged indexer cannot make the app unquittable. The stop is deliberately not cached: a macOS pause (`window-all-closed`) can be followed by `activate` → `startBackgroundResources`, and a later quit must stop the restarted singletons.
2. `stopBackgroundResources` closes the `~/.obelisk` watcher synchronously, before its first `await`. Part 1 alone was not enough (see verification): under continuous indexing churn `service.idle()` can outlast the 5s bound, and the bounded quit would still proceed with the obelisk watcher un-closed — the original crash. With the reorder, every watcher's `closed` flag is set before the function yields, so even a bound-cut quit leaves teardown-time deliveries hitting `if (closed) return`.

`createDeferredQuit` is extracted into its own module (same pattern as `file-reference.ts`) so the quit contract is testable in CI; logic is unchanged from the live-verified version.

## Verification

- [x] `npm test` — 720 tests, 712 pass / 8 fail. All 8 failures are in `tests/dsh-context-window-plugin.test.mjs` (rollover area), reproduce on untouched `main` in the same environment, and are an artifact of the local Node being v25.9.0 (Homebrew) while CI pins Node 22; the failing tests do not import the changed code. The CI runs on this PR are the authoritative gate.
- [x] `npm run lint` — 0 errors, 9 pre-existing warnings
- [x] `npm run test:electron:all` — 5/5 suites (macOS). Note: `electron-session-virtualization`'s "offscreen UUID focus" assertion is load-flaky in this environment — it failed 3 runs during a busy-machine window and passed 4 runs including with this change applied; the suite runs as a self-contained fixture main process and never loads the changed code
- [x] New tests actually run in CI (`.github/workflows/`): `tests/app-quit-teardown.test.mjs` (4 tests) and the new `close()` sync-guard contract test in `tests/adaptive-watcher.test.mjs` are picked up by the root `npm test` (`tests/*.test.mjs`) on all three platforms
- [x] Live quit-path repro, end to end from the real entry point: dev app under continuous watcher churn (writes to `~/.obelisk/quit-probe.md` every 50 ms plus live session-indexing churn), then a Cocoa `terminate` (the same path Cmd+Q takes):
  - with only the `before-quit` change: crash reproduced — same `napi_throw` fatal in `@parcel/watcher` during `CleanupHandles`
  - with the watcher-close reorder added: clean exits, ~7 s to quit (bounded path engaged under load) — 2/2 before the module extraction, 1/1 again on the final head with the extraction applied
- [x] Re-ran the checks above on the current head (`be7f808`)

Review follow-ups addressed in `be7f808`:

1. **adaptive-watcher contract test for the `close()` synchronous guard** — `tests/adaptive-watcher.test.mjs` now pins that an event delivered in the same tick as `close()`, with the unsubscribe still pending, is a no-op, and that the guard does not depend on `close()` settling. (The existing test only covered delivery after an awaited close.)
2. **CI-coverable behavior test for the deferred quit** — `tests/app-quit-teardown.test.mjs` covers: first request prevented and re-issued after the stop settles; re-entrant quit passing through without a double stop; a wedged stop bounded by `boundMs`; a rejected stop still ending in quit.

Remaining manual item from review: a human Cmd+Q during watcher activity. The programmatic equivalent above (Cocoa `terminate`, the same path) passed 3/3 under churn, but a maintainer Cmd+Q pass is still the last mile.

## Deliberately out of scope

- `indexer-service.stop()` fires its internal `watcher.close()` without awaiting the returned promise. The `closed` flag still flips synchronously, which is what teardown safety depends on; having the service await the close would be hardening, not a requirement of this fix.
- The 8 local `dsh-context-window-plugin` failures under Node 25.9.0 (unsupported version locally; CI pins 22) — untouched here.
- Windows/Linux quit (`window-all-closed` → `app.quit()` → the same `before-quit`): fixed by the same change, but only macOS was live-verified.
- The Windows CLI failure seen on `main` (`0276844`) was re-run and came back green — a runner flake, unrelated.

---

<details>
<summary><b>Main process / untrusted input</b></summary>

- [x] Transcript-derived path handling untouched; `shell.openPath` call sites unchanged
- [x] No file-read path added
- [x] No `innerHTML` or text-decoding change
- [x] No synchronous IO added — the change only reorders existing async cleanup and defers quit
- [x] Interception and navigation handling unchanged
- [x] New main source is `.ts` (ADR-0005): `app/src/main/quit-teardown.ts`

</details>


tommy0103:
Review follow-ups addressed in be7f808:

1. **adaptive-watcher `close()` sync-guard contract test** — pins that an event delivered in the same tick as `close()` (unsubscribe still pending, close unsettled) is a no-op; the guard does not depend on completion.
2. **deferred-quit behavior tests in CI** — `createDeferredQuit` extracted to `app/src/main/quit-teardown.ts` (logic unchanged) and covered by `tests/app-quit-teardown.test.mjs`: defer-until-settle, re-entrant pass-through, bounded wait, rejected stop.

Both run in the root `npm test`, so CI covers them on all three platforms. Re-verified on the new head: typecheck/lint clean, `npm test` 712/8 (same Node-25-only dsh failures as main), electron suites 5/5, live churn-quit 1/1 clean (3/3 cumulative).

Issue #176, opened 2026-09-18
[Linux] Electron window decoration causes duplicate title bar


FATFATHAO:
## Problem
On Linux the Electron window shows two title bars: the native one from the window manager, and an in-app title row rendered by the renderer. They show the same information, so the top of the window looks duplicated and wastes vertical space.

## Current behavior

<img width="2560" height="1440" alt="Image" src="https://github.com/user-attachments/assets/50a9100a-f9d4-464a-a2a1-043d01019bdb" />

`app/src/main/index.ts` passes `titleBarStyle: 'hiddenInset'` and `trafficLightPosition` unconditionally in `createWindow()`, and both are macOS-only. On Linux Electron ignores them and keeps native window decorations, while the renderer still draws its own title row (`app/src/renderer/src/App.vue`, `.titlebar` in `app/src/renderer/styles/base.css`) — so both appear. That row also keeps `padding-left: 78px`, the space reserved for the macOS traffic lights, which is empty here.

## Expected behavior

<img width="2560" height="1440" alt="Image" src="https://github.com/user-attachments/assets/7a74ebae-64ca-49a3-8799-7458cc1cfac7" />

Only one title bar. The app already renders its own title row with the window title, so I'd keep that one and make the window frameless on Linux (`frame: false`) instead of relying on native decorations. macOS keeps `titleBarStyle: 'hiddenInset'` and its traffic light position, unchanged.

One point I'd like a maintainer's call on: a frameless window has no native window controls. The in-app row is drag-only today — there are no close/minimize/maximize handlers in `app/src/preload` or the renderer — so closing or minimizing relies on compositor shortcuts unless we draw controls in-app. I can add custom controls as a follow-up, or leave the frameless window as-is if that is how the app should behave on Linux.

## Scope

- `app/src/main/index.ts` — platform-gated window options in `createWindow()`: `frame: false` on Linux, `titleBarStyle`/`trafficLightPosition` kept for the other platforms.
- No renderer changes; the in-app row stays as it is.
- Windows keeps today's behavior (the macOS-only options are ignored there, so it still shows native decorations plus the in-app row). I can extend the fix if you want it covered, but this issue is scoped to Linux.

## Reproducing

    cd app && npm run dev

    Observe the top of the window. The root `package.json` has no dev script, so the app must be started from `app/`.

[...]

PR #189, opened 2026-09-19
feat(app): add window controls on Linux

### LINKED ISSUE #183: [Linux] Double-click on the titlebar no longer toggles maximize

FATFATHAO:
## What and why

On Linux the window is frameless (#178) and no longer carries native decorations, so it has no minimize/maximize/close affordance at all, and the in-app titlebar lost double-click-to-maximize with the frame (#183). The only way left to control the window was a compositor shortcut.

This implements the three-part shape from #176, as the controls half of that follow-up:

- **Main** (`app/src/main/index.ts`) — one `win:control` handler with the `minimize` / `toggle-maximize` / `close` whitelist; the target is resolved with `BrowserWindow.fromWebContents(event.sender)` (the `capture:export` pattern, no renderer-supplied window id); `close` stays `win.close()`, so it keeps flowing through `window-all-closed → app.quit()`; `maximize` / `unmaximize` are pushed to the renderer on `obelisk:window-state` so the maximize/restore icon can toggle. An unknown action throws rather than being silently ignored.
- **Preload** — `windowControl(action)`, plus an `onWindowState` subscription shaped like the existing `onIndexUpdated` (same channel listener plus `removeListener` closure). `platform: process.platform` was the third item of that list and landed with the padding fix in #186, so it is not re-added here.
- **Renderer** (`app/src/renderer/src/App.vue`, `base.css`) — three buttons absolutely positioned at the right of `.titlebar`, rendered only on Linux, inheriting the blanket `-webkit-app-region: no-drag` rule. `.titlebar` also got a `@dblclick` → `toggle-maximize`, which is what #183 asks for (`.titlebar-text` is `pointer-events: none`, so the event lands on `.titlebar`, and the button strip stops the event so a double click on a button does not also toggle). The titlebar reserves the controls' footprint (`max-width: calc(100% - 130px)` = 3 × 40px buttons + 2 × 2px gaps + 6px right offset) so the window title stays centred on the window rather than being pushed left.

Before (no window controls) — after (minimize / maximize-restore / close):

Before: frameless Linux window with no window controls

After: the same window with minimize, maximize/restore and close in the titlebar

X11 (KWin) — minimize, maximize/restore and close, plus double-click:

https://github.com/user-attachments/assets/5884b4e1-9d04-4d36-9531-49ed15539dec

niri (Wayland) — the same controls; maximize/restore and close behave, minimize is inert (see below):

https://github.com/user-attachments/assets/35aa9864-fa9d-4388-a45b-32c35b8292a9

The terminal visible in the first recording is the `npm run dev` console: the `ui/gl/...` and `Exiting GPU process due to errors during initialization` lines are Electron's renderer GPU process failing to initialise ANGLE/EGL on this NVIDIA box and falling back — they appear on `main` in dev mode too and are unrelated to window options.

Closes #183

## Verification

All of the below was re-run on the current head (`6d4f886`), after rebasing onto `main` (`4a679b3`).

- [x] `npm test` — **736 pass / 0 fail**
- [x] `npm run lint` — 0 errors (9 pre-existing warnings, unchanged)
- [x] `npm run test:electron:all` — **2/5 suites** on this machine, with per-suite pass/fail counts identical to untouched `main` measured in the same session: `concurrency 15/0` · `images 11/1` · `virtualization 0/1` · `reader-state 6/5` · `file-references 30/3`. No Linux session reaches 5/5 here; on a real X11 session the same branch reaches 4/5 with the same per-suite deltas vs `main` (`images` passes with `--use-angle=swiftshader` while `virtualization` regresses with it) — the same matrix as #178.
- [x] New tests actually run in CI (`.github/workflows/cli.yml` runs `npm test` on Linux; both files below are picked up)
- [x] No existing assertion was loosened. The only change inside an existing test file is `on() {}` added to the 13 `FakeBrowserWindow` mocks in `tests/app-main-settings.test.mjs`, because `createWindow()` now registers `win.on('maximize' | 'unmaximize')`; no assertion, threshold or expectation was changed.
- [x] Every capability described above was exercised end to end from the outermost entry point — `npm run dev` on both sessions, clicking the real buttons (the two recordings above)

New tests:

- `tests/app-window-controls.test.mjs` — the preload bridge: all three actions go out on the single `win:control` channel, and `onWindowState` subscribes on `obelisk:window-state` and removes the same listener on unsubscribe.
- `tests/app-main-settings.test.mjs` — `the win:control IPC applies only whitelisted window actions to the sender window`: `minimize`, `toggle-maximize` twice (maximize, then unmaximize), `close` via `win.close()`, the window resolved from `event.sender`, and an unknown action throwing without touching the window.

Manual checks, both sessions:

| | minimize | maximize / restore | close |
| --- | --- | --- | --- |
| X11 (KWin) | window hides, comes back | button and double-click both toggle | exits |
| niri (Wayland) | inert — see below | button and double-click both toggle (niri honours a window-requested maximize) | exits |

## Deliberately out of scope

- **Minimize on Wayland is a platform limitation, not an omission.** Electron documents it on the `minimize` event: "On Wayland, 'minimized' is not currently a supported state." On niri the button is therefore inert — `win.minimize()` is accepted and nothing happens — while all three controls work on X11. I would rather ship the button and note this than special-case one action per session.
- **Double-click is gated on `process.platform === 'linux'`, not "non-macOS".** #183 says "Linux (or non-macOS)". Windows keeps its native decorations, so an app-level `toggle-maximize` would double-toggle there together with the native behaviour. If you would rather have the wider gate, that is a one-line change.
- **The padding fix and these controls landed as two PRs.** Your #183 note says the double-click half is "natural to land in the same PR" as the IPC, and #176 asks for the 78px allowance to be platform-gated "in the same pass". I split it: #186 (platform-gated padding, merged) and this PR (controls). The reason was revert surface — the padding change could land while this IPC was still in review — but if you would prefer them as one diff, I can fold this back into a single PR.
- **The title stays centred on the window** (the `max-width` guard above) rather than the right side being reserved by padding, so #184's result is preserved.
- **No automated click-the-button Electron UI test.** The new tests pin the IPC contract and the main-process handler; there is no Electron UI harness in CI (there is no Electron job at all) that could click a real titlebar. The evidence for the click path is the two recordings.
- **Wayland minimize also has no test.** The suites run the window in a headless-ish session where minimize is not meaningful, and there is nothing in-process to assert for a state the platform does not implement.

---

<details>
<summary><b>Renderer / Electron UI</b> — expand if you touched <code>app/src/renderer</code> or row rendering</summary>

- [x] Reader-anchor assertion added — **N/A here**: no row-rendering path is touched. `.titlebar` is a fixed 32px flex row and the controls are absolutely positioned inside it (`position: absolute; right: 6px`), so no row height, viewport height or scroll geometry changes.
- [x] Async media covered at mount, size-available, and load/error — **N/A**: no media element involved.
- [x] Tested both at rest and mid-scroll — **N/A**: no scroll behaviour touched.
- [x] No `loading="lazy"` on virtualized rows — not touched.
- [x] No hardcoded colors or type sizes in shadow DOM — the buttons use `var(--muted)`, `var(--surface-hi)`, `var(--fg-2)`, `var(--danger-soft)`, `var(--danger)`; the icons are `currentColor` and sized by CSS (12px), not by attributes inside the SVG.
- [x] One visual treatment per user-visible concept — maximize and restore are one button, one glyph at a time.
- [x] Library calling conventions either accept both shapes or are pinned in code — the icon constants are plain strings in `store.js`, the same convention as `FOLDER_SVG`.
- [x] Every renderer probe has a deadline and an `error` → reject path — the click path is a fire-and-forget `ipcRenderer.invoke` (nothing is awaited, nothing is measured), and `onWindowState` returns an unsubscribe that `onUnmounted` calls.
- [x] No assertion with sub-pixel headroom — the new tests assert behaviour, not pixels; the pixel numbers in this description are measurements, not test thresholds.

</details>

<details>

- [x] Transcript-derived paths do not reach `shell.openPath` unguarded — **N/A**: no transcript value reaches this path; the only input is a typed action string.
- [x] File reads triggered by transcript content are allowlisted — **N/A**.
- [x] No `innerHTML` / detached `<textarea>` decoding of untrusted text — the three icons are static in-repo constants (`store.js`) rendered with `v-html`; no runtime or transcript text passes through it.
- [x] No synchronous IO in the main process — the handler only calls `win.minimize()`, `win.maximize()`, `win.unmaximize()`, `win.close()`.
- [x] Interception scope not broadened beyond this feature — one new channel (`win:control`) plus the `obelisk:window-state` push, both scoped to the sender's window.
- [x] New main/preload sources are `.ts` (ADR-0005) — existing `.ts` files were extended; no new main/preload file.

</details>


tommy0103 (approved):
LGTM. Verified beyond CI:

- The handler matches the untrusted-input contract: window resolved from `event.sender`, action whitelisted by the switch with a throw on anything else — and the new main-process test pins exactly that, including ignoring a renderer-supplied window argument.
- Platform gating is right: controls render only on Linux, dblclick is Linux-only (double-toggle on Windows correctly avoided), `has-controls` absent means the #186 centering and the #184 padding result are untouched elsewhere. `win.on('maximize')` is a harmless no-op on macOS.
- Details done right: explicit `no-drag` on the controls container (the blanket rule only covers `button`), `@dblclick.stop` on the strip, `max-width: calc(100% - 130px)` math checks out, icons follow the `FOLDER_SVG` static-constant convention with no untrusted text.
- Test-count math reconciles (715 + 13 from #182 + 5 from #188 + 3 here = 736).
- Ran the branch on macOS (`npm run dev`): traffic lights in place, no buttons in the titlebar, title offset unchanged, double-click still native zoom — no macOS regression. Clean quit via the #188 path.

Non-blocking notes: Wayland minimize being inert is a documented platform limitation and shipping the button with the note is the right call; the "rebased onto main (4a679b3)" wording in the description is stale (actual base is bd7881a, mergeable); two cosmetic CSS spacing nits (`.titlebar-icon svg{`) that lint doesn't cover.

PR #195, opened 2026-09-22
fix(core): reject negative query limits

### LINKED ISSUE #152: Negative `limit` returns the entire index instead of being bounded
## Summary

List helpers accept a `limit`, but a **negative** `limit` is passed straight into SQL `LIMIT ?`. SQLite treats a negative LIMIT as "no upper bound", so `limit: -1` returns the entire index instead of erroring or falling back to the default.

Since these helpers exist to keep retrieval bounded, a single bad value can dump the whole corpus into an agent's context.

## Reproduction

Index used: 708 sessions / 25,397 messages.

```js
// obelisk --query <script>
return {
  search_neg:    search('obelisk', { limit: -5 }).length,   // 501
  search_pos:    search('obelisk', { limit: 5 }).length,    // 5
  search_zero:   search('obelisk', { limit: 0 }).length,    // 0
  search_dflt:   search('obelisk', {}).length,              // 20
  sessions_neg:  sessions({ limit: -1 }).length,            // 708  (all sessions)
  sessions_pos:  sessions({ limit: 1 }).length,             // 1
  recent_neg:    recent(-1).length,                         // 708
  failures_neg:  failures({ limit: -1 }).length,            // 400
  subagents_neg: subagents({ limit: -1 }).length,           // 268
};
```

`limit: -5` and `limit: -100` both return the same count, confirming the value is ignored rather than interpreted.

## Root cause

The default in the destructuring only applies to `undefined`, so a negative number flows through unchanged:

```ts
// packages/core/src/query.ts
const search = (text: string, opts: QueryOptions = {}) => {
  const {
    limit = 20,        // line 262 — applies to undefined only
    ...
  } = opts;
  ...
  stmt.all(matchText, ...filterParams, limit);   // bound directly into LIMIT ?
```

The same shape repeats across the helpers:

```
packages/core/src/query.ts:262   limit = 20    (search)
packages/core/src/query.ts:383   limit = 100
packages/core/src/query.ts:402   limit = 100
packages/core/src/query.ts:421   limit = 200
packages/core/src/query.ts:446   limit = 50
packages/core/src/query.ts:497   limit = 50
packages/core/src/query.ts:506   limit = 100
packages/core/src/query.ts:691   limit = 10000  (raw window)
packages/core/src/query.ts:724   limit = 50
```

Verified still present on `main` (checked `packages/core/src/query.ts` at the time of filing), not only in the published 0.2.5 build.

SQLite's documented behaviour for a negative LIMIT is "no upper bound", which is why nothing errors.

## Expected

Either reject the value or clamp it — a negative `limit` should not silently mean "unbounded". Clamping to the helper's default (`Math.max(1, limit)`) or throwing a validation error both seem reasonable; the choice is yours.

## Scope note

`limit: 0` and the default behave correctly, so this is specifically about negatives. `fileHistory` and `thread` return small counts here simply because the underlying sets are small; the same unbounded path applies.

## Environment

- `@obelisk-apps/cli` 0.2.5, macOS arm64
- Reproduced via both `--search` and `--query`

I'm happy to open a PR with the clamp if you'd like it, though I'd want to confirm which behaviour you prefer first.

22Goose:
## What and why

SQLite treats a negative bound parameter for `LIMIT ?` as no upper bound. The query helpers passed caller-provided negative limits through unchanged, so values such as `sessions({ limit: -1 })` could return the entire index and overflow an agent context.

Reject negative limits at the shared query API boundary before they reach SQLite. This covers every public helper with a dynamic limit, including `recent()` through `sessions()`, while preserving the existing `limit: 0` behavior. The authoritative API reference now documents the `RangeError` contract.

Reproduction before this change: `sessions({ limit: -1 })`, `recent(-1)`, and the other list helpers returned all matching rows. The root cause was SQLite's negative-`LIMIT` semantics combined with destructuring defaults that only handle `undefined`.

Closes #152

## Verification

- [x] `npm test` — green in CI on ubuntu/macos/windows; the Linux full-suite run reports **759 tests / 759 pass / 0 fail** ([run 35689190381](https://github.com/tommy0103/obelisk/actions/runs/35689190381)). The author's Windows checkout has 16 unrelated platform/environment failures (POSIX path assertions, symlink permissions, nested `npm` resolution, release-staging `.git` expectations).
- [x] `npm run lint` — 0 errors (11 existing warnings outside this change)
- [x] New tests actually run in CI (`.github/workflows/cli.yml` runs both `query.test.mjs` and `runtime-cli-envelope.test.mjs`)
- [x] Every capability described above was exercised end to end from `obelisk --query`
- [x] Re-ran the checks above on commit `712806b`

Focused results:

- `tests/query.test.mjs` — 33 pass / 0 fail
- `tests/runtime-cli-envelope.test.mjs` — 8 pass / 0 fail

## Deliberately out of scope

- `thread()` is unchanged because the current helper and authoritative API do not accept a `limit` parameter.
- No upper cap is introduced; this bug is specifically the accidental unbounded meaning of negative values.

## Maintainer amendments (`026708a`)

Pushed to this branch during review; re-verified locally (query 33/33, envelope 8/8, tsc + eslint clean):

- `raw()` now also rejects a negative `offset` (`raw() offset must be non-negative …`): it feeds `record.text.slice()`, where JS silently treats a negative start as "from the end of the text" — same silent-misinterpretation class as the negative `LIMIT` this PR fixes.
- `RangeError` messages now include the offending value, e.g. `sessions() limit must be non-negative (got -1)`. `--query` callers only ever see the error string, so the value is their only diagnostic context for a bad limit.
- Dropped the unused default `label` parameter so every call site names its helper; the api-reference now notes that the message names the helper and the offending value.


tommy0103:
Hi @22Goose — nice catch and clean fix. During review I pushed one follow-up commit (`026708a`) to this branch: it also guards a negative `raw()` `offset` (it feeds `text.slice()`, where a negative start silently reads from the end), includes the offending value in the `RangeError` message (`sessions() limit must be non-negative (got -1)`) since `--query` callers only see the error string, and drops the unused default `label` param. Tests were extended accordingly; everything re-verified locally and CI is re-running on the new head. Also ticked the `npm test` checkbox with the CI numbers. Speak up if you'd prefer a different message format.

22Goose:
Looks good to me. Thanks for the follow-up!

PR #197, opened 2026-09-22
feat(providers): add a Hermes Agent transcript source

### LINKED ISSUE #196: Add Hermes Agent (Nous Research) provider
## Proposal

Add a `hermes` provider adapter that indexes Hermes Agent (Nous Research) session history alongside the existing claude / codex / copilot / deepseek / kimi / omp / pi providers.

## Source format

Hermes moved from per-session JSONL to one SQLite store per install (per profile), WAL mode:

- Store `~/.hermes/state.db` (`$HERMES_HOME/state.db`; `%LOCALAPPDATA%/hermes` on Windows), plus `~/.hermes/profiles/<name>/state.db` for named profiles. `~/.hermes/sessions/*.jsonl` is retired upstream and no longer written or read.
- `sessions`: `id`, `source` (`cli` / `subagent`), `model`, `title`, `parent_session_id`, `started_at` / `ended_at` (REAL epoch), `message_count`, `tool_call_count`, `cwd`, `git_branch`, `profile_name`, `archived`, `hidden`, `session_key`, `chat_id`, `chat_type`, `thread_id`, per-session `input_tokens` / `output_tokens` / cache / reasoning columns (58 columns).
- `messages`: `id` (**INTEGER PRIMARY KEY AUTOINCREMENT**), `role` (user / assistant / tool), `content`, `tool_calls` (JSON), `tool_call_id`, `tool_name`, `timestamp` (REAL epoch), `reasoning` / `reasoning_content`, `active`, `compacted`, `display_kind` (24 columns).
- Sidecars `state.db-wal` / `state.db-shm`; the three files are one image. The layout is documented upstream (`developer-guide/session-storage`) and I also checked it against the host's own source.

From my local store (`schema_version` 30, ~259 MB plus a ~9 MB `-wal`): ~122 sessions (~106 `cli`, ~16 `subagent`, every `subagent` one carrying `parent_session_id`), ~16.6k messages, ~8.5k tool results, ~3.2k `active = 0` rows. Per-message `token_count` is 100% NULL; the per-session token columns are populated.

## Design highlights

- **Reader seam**: `hermes.ts` over `SqliteDb` / `SqliteRow` plus a small `hermes-node.ts` opener, injected through `builtins.ts` the way the Copilot chronicle is (ADR-0005). Every connection is read-only.
- **Identity**: `hermes:<encodeURIComponent(raw id)>:<sha256('hermes-store-v1\0' + profile + '\0' + normalized store path)>` — Hermes ids are install-global, so the profile separates named profiles and the store term keeps a copied or migrated `state.db` from colliding with the original.
- **Cursor**: `${max(state.db, state.db-wal) mtime}:0:hermes-snapshot-v1:${sha256(session fingerprint)}`, where the fingerprint is the mutable session fields plus a message watermark; `parse()` re-derives it and refuses a session that changed after discovery.
- **Watch targets**: `state.db`, `state.db-wal`, and the `profiles` tree.
- **Supersession**: `active = 0` → `inactive`; any non-empty `display_kind` → `hidden` (upstream treat those rows as operational traffic — delegation and notification bookkeeping — rather than user intent).
- **Subagent lineage**: a `source='subagent'` session is linked to its parent in `subagents`, with `total_tokens` from the session token columns. Upstream does not record the spawning call id, so `parent_tool_use_id` is null; unlike Claude Code's sidechain units the child stays a first-class session, because upstream stores it as a full one.
- **Failure handling**: a missing store, an unreadable `profiles/` directory, an unopenable database, or `schema_version > 30` is reported as an incomplete inventory and never retracts sessions. `ENOENT` on `profiles/` is normal.

## Out of scope (phase 2, separate issues)

- **Compaction-archived rows** (`active = 0 AND compacted = 1`): stored as `inactive`, matching upstream's default live read. Its display reads would include them and deduplicate repeated generations.
- **Gateway / messaging sessions** (`session_key`, `chat_id`, `chat_type`, `thread_id`, usually no `cwd`): read but not projected — the canonical `sessions` table has no column for them.
- **Query helpers that join subagents on `messages.agent_id`**: Hermes (like DeepSeek) never sets that column, so `messageCount` and the time filters of `subagents()` see no messages for these rows.

## Status

The adapter is written locally: `npm test` 769/769, typecheck and lint clean, `test:electron:all` at the same per-suite result as untouched `main`, and `maintainer-skills/adapter-review/SKILL.md` run in a clean session with no blockers left. Happy to send it as a PR, with a synthetic-schema fixture rather than my real history.

FATFATHAO:
## What and why

Adds Hermes Agent (Nous Research) as a transcript source, reading `~/.hermes/state.db` (or `$HERMES_HOME/state.db`, plus every `profiles/*/state.db`).

Hermes keeps one SQLite database per install (per profile) instead of per-session files, so it cannot be indexed by pointing a root at a directory. The adapter follows the Copilot chronicle seam: a pure `hermes.ts` over `SqliteDb`/`SqliteRow` plus a `hermes-node.ts` opener, with the Electron app injecting `better-sqlite3` and the CLI injecting `node:sqlite` (ADR-0005).

Mapping decisions worth reviewing:

- **Identity** is `hermes:<raw id>:<sha256('hermes-store-v1\0' + profile + '\0' + normalized store path)>`. Hermes ids are install-global, so the profile keeps named profiles apart and the store term keeps a copied or migrated `state.db` from colliding with the original.
- **Cursor** is `${max(state.db, state.db-wal) mtime}:0:hermes-snapshot-v1:${sha256(session fingerprint)}`, where the fingerprint is the mutable session fields plus a message watermark. `parse()` re-derives it and refuses a session that changed after discovery.
- **Supersession**: `active = 0` → `inactive`; any non-empty `display_kind` → `hidden` (upstream treats those rows as operational traffic, not user intent).
- **Subagent lineage**: a `source='subagent'` session is recorded in `subagents` as a link to its parent (`agent_id` = the child session, `session_id` = the parent), with `total_tokens` from the session-level token columns. `parent_tool_use_id` stays null because upstream does not record the spawning call id; unlike Claude Code's sidechain units the child's messages stay visible, since upstream stores it as a full session. Compression-triggered `cli` splits with a parent are not mapped to `subagents`.
- **Incomplete inventories never retract**: a missing store, an unreadable `profiles/` directory, an unsupported `schema_version`, or an unopenable database is reported through `reportIncompleteInventory` and produces no tombstones. `ENOENT` on `profiles/` is normal; anything else is not.
- `indexVersionMarker` bumped to `__hermes_canonical_transcript_v2__` (the uuid carries the store's INTEGER message id).

Closes #196

## Verification

- [x] `npm test` — **769 pass / 0 fail** (759 pre-existing + 10 new in `tests/hermes-provider.test.mjs`, plus the registry/app expectations extended)
- [x] `npm run lint` — 0 errors / 11 warnings (9 pre-existing, 2 from the gitignored `packages/dsh-plugin/dist/*.d.ts` build output)
- [x] `npm run test:electron:all` — **2/5 suites**, identical to untouched `main` on this machine, re-run on the rebased head: `images` 1 failure, `virtualization` 1, `reader-state` 5 (`concurrency` and `file-references` pass). The three failures are pre-existing renderer/Wayland issues (`'--ozone-platform=wayland' is not compatible with Vulkan`), unrelated to this change.
- [x] New tests actually run in CI — `tests/*.test.mjs` is the workflow glob, and `tests/hermes-provider.test.mjs` matches it
- [x] No existing assertion was loosened — two expected lists gained entries (`tests/provider-registry.test.mjs` catalog + watch targets, `tests/app-main-settings.test.mjs` main-process watch targets, both with `HERMES_HOME` pinned so the fixture is hermetic)
- [x] Exercised end to end from the outermost entry point — `obelisk --build` against a synthetic store indexed both sessions (`source='hermes'`, tool call + result, a `display_kind` row stored as `hidden`, and a `subagents` row linking the delegated child to its parent); the real store indexes 120 sessions / ~16.3k messages in ~8 ms per discover with 0 parse errors
- [x] Re-ran the checks above on the current head after the most recent change

## Deliberately out of scope

- **Compaction-archived history** (`active = 0 AND compacted = 1`, ~3.2k rows): stored as `inactive`, matching upstream's default live read. Showing it would mean adopting upstream's display semantics *and* its `_dedupe_display_generations` (11 duplicate groups in my store); no existing provider does that today, so I kept the conservative mapping.
- **Gateway session metadata** (`session_key`, `chat_id`, `chat_type`, `thread_id`): read but not projected — the canonical `sessions` table has no column for it and no provider maps it today.
- **Usage attribution** (`messages.token_count` is 100% NULL; `session_model_usage` is per-model aggregate): projected as `null` rather than invented, the same position as the Copilot chronicle.
- **Query helpers that join subagents on `messages.agent_id`**: Hermes (like DeepSeek) never sets that column, because a delegated session is stored as its own session rather than as sidechain messages inside the parent. The `subagents` row is written and `source` filters return it, but `messageCount` and the time filters of the `subagents()` helper see no messages — worth a follow-up if you want those surfaces to cover non-Claude-style subagents.
- **Subagent labels**: upstream usually leaves a delegated session's `title` empty, so `description` is often null, and `duration_ms` is omitted when the recorded clocks regress rather than reported as a negative span.
- **FTS tables** (`messages_fts*`): not read; Obelisk builds its own index.

---

<details>

- [x] Read all provider implementations in `packages/core/src/providers/` in full first
- [x] Session identity is composite — `(raw id, profile, store path)`, see above; a test covers the same host id under two profiles and two stores in different directories
- [x] A test actually calls `discover()` against each supported directory layout (default store, named profile, WAL pair, unsupported schema, deleted session, unreadable `profiles/`)
- [x] Directory layout verified against upstream source/format docs (`developer-guide/session-storage`, plus the host's own `hermes_constants.py` / `hermes_state_common.py` for `SCHEMA_VERSION`, the `profiles/<name>` layout, and `.deleted`)
- [x] Canonical transcript invariant holds (ADR-0007) — the persist round-trip test writes a Hermes session through `persist()` and replays it without duplicating rows
- [x] Text-less records still emit a row (`content` NULL/empty and tool-only rows produce `text: null` with `content_type` `tool_use`/`unknown`)
- [x] `indexVersionMarker` bumped (`v1` → `v2`) because the uuid now carries the store's INTEGER id instead of a synthetic one
- [x] "Should not be displayed" uses `visibility`; `is_sidechain` keeps its single meaning
- [ ] Cursor detects same-millisecond rewrites (mtime + ctime + size + inode) — **not literally**: this cursor is `max(state.db, state.db-wal)` mtime plus a *content* fingerprint (session fields + message watermark), the same shape as Copilot's chronicle cursor. A same-millisecond rewrite that leaves the fingerprint unchanged is not detected; the source is a live WAL database whose `-wal` mtime moves on every write, so I judged the content digest the stronger signal. Flagging it rather than silently diverging.
- [x] Unknown/newer versions are skipped and recorded, not thrown on (`schema_version > 30` → one inventory issue, no tombstones)

</details>

<details>

- [x] No transcript-derived path reaches `shell.openPath` — the adapter only reads the store it was pointed at
- [x] No new file reads triggered by transcript content (the adapter reads one SQLite file per store)
- [x] No `innerHTML` / decoding of untrusted text in the main process
- [x] No synchronous IO beyond what the existing Copilot chronicle opener already does (read-only `DatabaseSync`/`better-sqlite3` opens of the same shape)
- [x] Interception scope not broadened
- [x] New main sources are `.ts` (ADR-0005); the opener is injected, so core never imports a driver

</details>

<details>
<summary><b>Indexing / daemon</b> — expand if you touched <code>indexer.ts</code>, <code>provider-indexing.ts</code>, or write coordination</summary>

- [x] Read-only while the daemon heartbeat is fresh — the provider only ever opens stores read-only and never writes the host's database
- [x] Live refresh has a real refresh point: `state.db`, `state.db-wal`, and `profiles/` are all declared as watch targets
- [x] Configuration mismatch is explicit: an unreadable store is reported as an incomplete inventory instead of silently falling back to a stale index
- [x] Full rebuild vs. trigger: unchanged from every other provider (same `discover()` contract, same marker/replay machinery)

</details>


tommy0103 (commented):
Thanks for adding the Hermes source. I reviewed `6e76f58` against #196 and the upstream storage format. I found a few correctness issues that should be addressed before merge:

1. `pathInside()` (hermes.ts:101-105) uses a literal `/` after `path.normalize()`. On Windows, an indexed `C:\...\state.db#session:...` is not recognized as being inside its Hermes home. If that home becomes unavailable, `discover()` does not report an incomplete inventory and the final loop emits tombstones for its sessions. Please use a platform-aware containment check and cover this path on Windows.
2. `storePaths()` (hermes.ts:204-214) silently skips any profile entry whose `statSync()` fails, then returns `error: null`. `readdir(profiles/)` can succeed while `stat(profiles/name)` returns `EACCES`; in that case existing sessions for the profile are retracted as if deleted. Please report the inventory as incomplete for access/I/O errors, while allowing a genuinely removed entry to retract.
3. The message query at hermes.ts:388-394 never reads `codex_message_items`. Hermes's session-storage docs say a final Responses reply may live only in that column with empty `content`, so this adapter indexes that reply as a textless message. Please project the final reply when canonical `content` is empty, with a fixture from that format.
4. `readToolCalls()` (hermes.ts:348-359) chooses `id` and ignores `call_id`. Hermes uses `call_id || id` for Codex Responses calls, where the two IDs can differ and tool rows refer to `call_id`. The resulting tool call and result get different canonical IDs; `raw()` also does not search `$.call_id`. Please normalize this ID consistently across calls, results, and raw lookup.
5. The cursor combines max(db/WAL mtime) with message count/maxima (hermes.ts:158-166, 243-275). An in-place update to an existing message's content, tool calls, or visibility leaves those maxima unchanged; a same-millisecond write can therefore be skipped indefinitely. This also misses the mtime+ctime+size+inode cursor requirement in `CONTRIBUTING.md`.

The new tests use a handwritten `messages.id TEXT PRIMARY KEY` fixture, although Hermes stores an integer autoincrement ID, and they do not assert that direct `assembleSessionDetail()` equals assembly after a SQLite round-trip. Please add real-format coverage for the cases above and the ADR-0007 canonical transcript invariant.

FATFATHAO:
Thanks for the careful review, and sorry it took a round trip to get the Windows path right. Everything you listed is addressed — `ae1c421` on top of `b8a9823`:

**1. `pathInside()`** — no more literal `/`: containment goes through `path.relative()` with the platform's case rules. The path operations are now injected (`ContainmentPathOps`), so the win32 branch is covered from a POSIX host as well; the tests check a store under its home, a named profile store, a case-different path, and a sibling whose name merely starts with the home (`hermes-other`). The suite sits in the CI list that already runs on every platform.

**2. `storePaths()`** — returns `{ stores, errors }` now. `ENOENT` stays silent, so a genuinely removed entry still retracts, while any other failure reports an incomplete inventory, which suppresses retraction.

**3. `codex_message_items`** — projected when `content` is empty: assistant `message` items, `output_text`/`text` parts, joined with newlines, mirroring `_replay_message_items`. A row with nothing projectable still yields `text: null`. For context: none of the ~16.7k message rows in my own store populate that column, so this coverage is synthetic — it follows the documented shape and upstream's decode rather than an observed row.

**4. `call_id`** — calls use `call_id ?? id`, and `raw()` searches `$.call_id` before `$.id`. On the blast radius: all 8,877 calls in my store have `call_id == id`, so this changes canonical ids only for Responses-shaped sessions.

**5. Cursor** — now mtime + ctime + size + inode for both `state.db` and `state.db-wal`, plus a wider per-session watermark (content / tool_calls byte sums, active, compacted, display_kind). The regression test pins the store's mtime to a whole second and swaps in a same-length replacement, so only the ctime/inode part of the signature can notice the edit. The marker moves to v4 so existing indexes replay, and a new test asserts that an index written under `__hermes_canonical_transcript_v3__` is replayed instead of trusted.

**Fixture and ADR-0007** — the fixture tables now use Hermes' own definitions (`id INTEGER PRIMARY KEY AUTOINCREMENT`, `codex_message_items`, and the neighboring columns), and a new test asserts that direct `assembleSessionDetail()` equals assembly after a SQLite round-trip.

Numbers on `ae1c421`: `npm test` 778/778, `tests/hermes-provider.test.mjs` 19/19, typecheck clean, `npm run lint` 0 errors (11 pre-existing warnings). A read-only pass over my own store (122 sessions, 15,139 messages) parses with 0 errors, resolves `raw()` for 400/400 sampled messages and 400/400 tool calls, and finds no canonical tool id shared between two sessions.

One unrelated note while I'm here: the Windows job's failure is not from this branch. `main` fails the same `codex strict reconcile heals a touched-but-unchanged source` assertion (run [35719388678](https://github.com/tommy0103/obelisk/actions/runs/35719388678)) with `expected` and `actual` printed identically, which looks like sub-millisecond float precision on Windows rather than anything this adapter touches.

其实是古法手写时候忘了Windows了（，刚让codex给我扫了一遍，把提到的地方都修改了一下

tommy0103 (commented):
Thanks for the follow-up. I checked `ae1c421`: the Windows containment, unreadable-profile, Responses text, `call_id`, cursor-signature, and ADR-0007 test fixes are present, and the three current CI jobs pass. I still see three issues:

1. `cursorFor()` (hermes.ts:232-237) puts the **store-wide** `state.db`/WAL signature into **every session's** cursor. A write to one session therefore changes every cursor; `discover()` (lines 789-816) schedules a retract-and-full-reparse for every historical session. `changedPaths` also invalidates every session under the home. With the reported 122 sessions / 15k messages, one live append can replay the whole store. Please add a two-session regression asserting that changing one leaves the other's unit unscheduled, and measure the full refresh cost before choosing the cursor/invalidation scheme.
2. In `storePaths()` (lines 260-276), `ENOENT` from `statSync(profileDir)` after `readdirSync(profilesDir)` already listed that entry is an enumeration/stat race, not proof of a completed deletion. The code leaves `inventoryComplete` true, so lines 822-839 can tombstone that profile's indexed sessions during a move/replacement. ADR-0001 says enumeration/stat races should be treated as incomplete inventory and retried; a later stable scan can then confirm deletion.
3. `parse()` checks `sessionFingerprint()` (line 490), then reads messages in a separate SELECT (lines 496-502), without a read transaction or post-read validation. In WAL mode a writer can commit between those statements, so the parser can emit old session metadata with new messages and return the cursor from discovery. Please read one consistent snapshot, or reject a changed snapshot before publishing the cursor. See SQLite's snapshot-isolation documentation: https://www.sqlite.org/isolation.html.

The PR description also still reports marker `v2` and 769 tests, while this head uses `v4` and reports 778 tests; please bring those claims up to date. The Responses fixture remains synthetic, so it does not yet satisfy `CONTRIBUTING.md`'s real-provider-output fixture rule.

tommy0103:
@FATFATHAO Heads-up: #200 has merged into `main` (`87ac8cb`). `app/src/main/indexer-service.ts` now forwards invalidations for provider-declared exact `kind: 'file'` targets regardless of suffix, so the root Hermes `state.db` and `state.db-wal` targets can reach changed-path indexing.

I noticed one remaining coverage gap for #197: databases inside `profiles/` are covered by its `tree` target, not exact file targets. The App still filters ordinary existing `.db` and `.db-wal` tree events, so writes to an existing named profile would wait for periodic full reconciliation. Please address or explicitly account for that path, including profiles created after watcher startup, and pin the intended behavior with an App watcher/indexer test when you bring the new `main` into your branch.

FATFATHAO:
Thanks — everything from all three rounds is on the head now. I rebased onto `87ac8cb` from #200 and squashed the branch into three commits, so the SHAs you looked at are gone; the diff is the shape you reviewed plus the fixes below.

From your round-2 list: invalidation is per session now — a store gate decides whether the file is opened at all, a per-session cursor decides which sessions are re-read. One appended message went from 123 scheduled units to 1, and an idle store is not opened at all. The `ENOENT` that follows `readdir` is treated as the enumeration race `docs/adr/0001` describes: it reports an incomplete inventory and retracts nothing. A store that is merely missing is not a deletion either — the host quarantines and replaces `state.db`, so the first scan only records a missing marker and the second one retracts. The fingerprint check and the message read now share one read transaction.

For #200: profile stores are declared as exact file targets (`profiles/<name>/state.db` and its `-wal`), `profiles/` stays a tree target, and there is an App-side test that drives the real `watchTargets()` output through the indexer. A profile created while the watcher is already running is picked up on its directory event.

Two more things came out of running the adapter against my real store (127 sessions), both ADR-0007 round-trip breaks that fixtures alone did not catch: a message with more than one tool call came back in the order the `(session_id, name)` index imposes instead of source order (68 of the 126 sessions), and truncation landing on an astral character left a lone surrogate, which SQLite rewrites to U+FFFD.

The description is updated with the numbers re-run on this head: 810 pass (49 provider cases), the marker is `v5`, and `display_kind` is described correctly now — only `hidden` is hidden, and the `async_delegation_complete` rows stay visible.

tommy0103:
@FATFATHAO 感谢你把 Hermes provider 做到可合并的状态，尤其是命名 profile、SQLite/WAL 增量索引、真实数据 fixture 和跨平台 watcher 的细节。辛苦了！

合并前，我们在这个 PR 分支上补了最后几处修复和回归用例：

- `state.db` 暂时缺失或不可读时保留已索引会话，数据库恢复后再依据完整清单处理真正删除的会话；整个 profile 目录消失则仍视为删除。
- 为 SQLite 投影摘要加入类型和长度边界，并将 canonical marker 升到 v7，避免不同字段拼接后发生摘要碰撞而留下旧索引；修复 Hermes home 路径包含 `#` 时的 raw source 定位。
- 让 `rawHermes` 跳过格式错误的 `tool_calls` 数据，继续查找后续有效调用，并补上对应测试。

最新提交 `c4474b3` 已通过 `npm test`（840/840）、typecheck、lint（0 errors）和 Electron 测试（5/5）；Ubuntu、macOS、Windows 的 CI 也全部通过。感谢你耐心处理多轮 review！

PR #198, opened 2026-09-22
feat(core): add ZCode provider

### LINKED ISSUE #180: Add ZCode (Z.ai) provider
## Proposal

I'd like ZCode supported as an Obelisk provider, so its history shows up alongside other agents.

ZCode stores session history in a SQLite database rather than JSONL transcripts:

- `~/.zcode/cli/db/db.sqlite`: the session store
- `~/.zcode/cli/rollout/model-io-*.jsonl`: model request/response logs, not the canonical transcript

## What the history contains

From my local ZCode history (CLI 0.16.5):

- 11 sessions (6 interactive plus 5 subagent children), with parent links between them
- 135 messages, each with role, parent-message chain, per-message model, and token counts including cache reads/writes; message content is split into parts
- 169 tool calls with timing and error details

Two caveats: ZCode is closed-source with no public format documentation, so the layout above comes from inspecting my local database; and my history contains no compaction, so that part is unverified.

Happy to share the full schema or a sample database if that helps.

tsx8:
[#180 — Add ZCode (Z.ai) provider](https://github.com/tommy0103/obelisk/issues/180) | [fixture sanitization contract](tests/fixtures/zcode/README.md)

## Why the change

ZCode keeps its whole history in one SQLite database (`~/.zcode/cli/db/db.sqlite`) instead of JSONL transcripts, so this PR adds a read-only SQLite provider that projects its sessions (messages, reasoning, tool calls, rewind/compaction supersession, subagent children) into the shared index alongside every other agent.

Closes #180

## Special things to note

- Change detection is content-based, not watermark-based: ZCode's `touchSession` is `max(time_updated, ?)` and user-message re-saves pass the original creation time, so watermarks can fail to advance on real writes. Every storage mutation (append, in-place re-save, delete, rewind, compaction) is handled by one strategy: atomically replace the whole session (discovery attaches `retractSessionIds`, parse re-emits with `countMode: 'total'`).
- Model identity follows a shape-authority rule for the migration-0020 rename (`{providerID, modelID}` → `{providerId, modelId}`): if any member of the new shape is present, the new shape owns the whole selection; a parse failure yields a `null` model, never a per-field fallback to old keys.
- The test corpus is a sanitized aggregate of twelve real-model scenario runs (GLM-5.3, zcode CLI 0.16.9, schema 0022), not a hand-written fixture; the byte-level sanitization gate and the aggregation contract are documented in `tests/fixtures/zcode/README.md`.

## Change outline

```diff
 packages/core/src/providers/
+├── zcode.ts                    # ZCode adapter: read-only SQLite, full-reparse
 ├── builtins.ts                 # + createZcodeProvider under root key 'zcode'
 tests/
+├── app-zcode-index.test.mjs    # 13 tests: e2e index/search, incremental, tombstone,
+│                               #   rewind/compaction inactive, dual-shape models,
+│                               #   subagents, changedPaths routing, ADR-0007 gate
+└── fixtures/zcode/             # sanitized real-model aggregate + sanitization contract
 ├── provider-registry.test.mjs  # + catalog entry and db/-wal watch targets
 ├── provider-inventory.test.mjs # + unusable/missing-db inventory semantics
 └── app-main-settings.test.mjs  # + db.sqlite & db.sqlite-wal watch targets
```

`README.md`, `skill-doc/`, and `.github/workflows/cli.yml` list the new source and its CI test.

Discovery and parse (all source reads go through one read-only, WAL-aware connection):

```text
discover(root = ~/.zcode/cli)
  db = <root>/db/db.sqlite            (watched with its -wal sidecar)
  missing / inaccessible / no transcript tables
    → reportIncompleteInventory, zero units, zero tombstones
  one consistent snapshot:
    per session row:
      L1  session.time_updated + message count      cheap batch compare
      L2  sha256 over raw session/message/part rows catches L1-invisible rewrites
      changed → unit{ key `${db}#z:${rawId}`,
                      sessionId `zcode:<12-hex sha256(realpath(db))>:<rawId>`,
                      retractSessionIds: [sessionId] }  replace, never merge
    indexed-but-gone in this db → tombstone unit (retract-only; cursor is a
      'tombstone' sentinel that can never match a real fingerprint, so a
      byte-identical restore is re-indexed)

parse(unit)  — full reparse in one transaction; session vanished → throw,
               the unit is skipped and the last-good index survives
  semantics triad (ui × provider × transcript) → visible / visible+is_meta /
    hidden, unknown combinations fail closed to hidden
  rewind: outside keptMessageIDs ∪ post-branch-cut → inactive
  compaction: messages up to the inclusive tail_start_id → inactive
  emits text/thinking messages, tool_call + tool_result, compact summaries,
    and the subagent record from the child's own task_type + parent link
  cursor `${time_updated}:${messageCount}:v2:${fingerprint24}`
```

Superseded history stays retrievable via `includeInactive: true`; session identity binds the database's realpath, so a database recreated at the same path retracts stale snapshots instead of mixing histories.

## Verification

- [x] `npm test` — 774 tests / 753 pass / 21 fail. All 21 failures are in `kimi-manifest.test.mjs` and reproduce identically on a clean tree at the base commit `b4c20eb` (environment-attributed, unrelated to this change).
- [x] `npm run lint` — 0 errors (11 pre-existing warnings)
- [ ] `npm run test:electron:all` — not run; no `app/` sources changed (only `tests/` expectations)
- [x] New tests actually run in CI — `tests/app-zcode-index.test.mjs` is covered by the Linux `npm test` glob and added to the explicit macOS/Windows list in `.github/workflows/cli.yml`
- [x] No existing assertion was loosened — touched tests only gained `zcode` entries in existing expectation lists
- [x] Every capability described above was exercised end to end from the outermost entry point — `app-zcode-index.test.mjs` runs the real fixture through `buildIndex` (app indexer) and a full-stack CLI `--query` search
- [x] Re-ran the checks above on the current head after the most recent merge — numbers are from this tree at `b4c20eb`, the current `origin/main` head

Adapter review: run in a clean session per CONTRIBUTING (`maintainer-skills/adapter-review`, not the authoring session). 1 blocking + 6 major findings, all fixed, each pinned by a mutation-verification regression test.

<details>

- [x] Read all provider implementations in `packages/core/src/providers/` in full first
- [x] Session identity is composite, not the source id alone — `zcode:<12-hex sha256(realpath(db.sqlite))>:<raw session id>`
- [x] A test actually calls `discover()` against each supported directory layout — single layout `<root>/db/db.sqlite`; called directly in `provider-inventory.test.mjs` and via `buildIndex`
- [x] Directory layout verified against upstream source or format docs — layout and write behavior verified against the zcode CLI source (`touchSession`, messages repository ordering), cited in `zcode.ts` header comments
- [x] Canonical transcript invariant holds: direct assembly == SQLite round-trip (ADR-0007) — pinned by a dedicated 17-session test
- [x] Text-less records still emit a row — file attachments and hidden transport-only messages keep a `text: null` message row
- [x] `indexVersionMarker` — first release, `__zcode_canonical_transcript_v1__`; no stored-row shape of an existing provider changed
- [x] "Should not be displayed" uses `visibility`, not a new meaning for `is_sidechain`
- [x] Cursor detects same-millisecond rewrites — content sha256 over the raw rows, strictly stronger than stat fields
- [x] Unknown/newer versions are skipped and recorded, not thrown on — unknown semantics kinds fail closed to `hidden`; unknown model shapes yield a `null` model; an unusable database reports incomplete inventory without poisoning other units

</details>

## Deliberately out of scope

- ZCode's legacy script-workflow metadata is not indexed (no fixture coverage yet); workflow child sessions appear as ordinary sessions. Follow-up.
- Subagent pairing emits no `parent_tool_use_id` — the parent link is `session.parent_id` only; no heuristic pairing until an explicit upstream bridge exists.
- `~/.zcode/cli/rollout/model-io-*.jsonl` request/response logs are not indexed (the session store is the canonical transcript).
- Moving the database to a different root leaves old sessions in place until a forced rebuild; only a database recreated at the same path self-retracts.
- Fixture gaps recorded in `tests/fixtures/zcode/README.md`: cross-compaction double rewind, fork-of-compacted, workflow scenarios.


tommy0103:
@tsx8 Heads-up: #200 has merged into `main` (`87ac8cb`). `app/src/main/indexer-service.ts` now forwards invalidations for provider-declared `kind: 'file'` watch targets regardless of suffix. Your ZCode `db.sqlite` and `db.sqlite-wal` targets already use that contract, so their changes can reach `buildIndex({ changedPaths })`; the adaptive watcher keeps them in its pinned poller.

Please bring the new `main` into #198 and rerun verification before landing, especially the App path from the watcher callback to changed-path indexing. No provider-side suffix workaround is needed.

tsx8:
@tommy0103 Merged `87ac8cb` into #198 (`a3e916a`) and added a real-fixture test for `WAL → watcher → changedPaths → index`. On Node 22, `npm test` passed 775/775; typecheck and lint passed. Electron passed 3/5 suites on this Nix host; the same two failed on clean `87ac8cb` under the same environment.

tommy0103 (commented):
Reviewing `a3e916a` against `main` (`87ac8cb`). The new real-fixture WAL → watcher → `changedPaths` → index test addresses the earlier live-refresh concern. I found the following remaining issues.

**Please address before merging**

1. **[P1] Cache reads are counted twice in usage.** [`totalInputTokens()`](https://github.com/tommy0103/obelisk/blob/a3e916aea7843796d21a74ea15cb471bc367f9c8/packages/core/src/providers/zcode.ts#L456-L467) adds `tokens.cache.read` to `tokens.input`. In the committed source fixture, one assistant message has `input=34872`, `output=81`, `total=34953`, and `cache.read=31808`: `input` already includes the cached tokens. This code stores `input_tokens=66680` for that message, exceeding the source's entire `total`. Please follow the source accounting and add an assertion against real fixture usage totals.

2. **[P1] Subagent sessions open with empty details.** [Child messages](https://github.com/tommy0103/obelisk/blob/a3e916aea7843796d21a74ea15cb471bc367f9c8/packages/core/src/providers/zcode.ts#L743-L764) all receive `agent_id=childSessionId`, while [`assembleSessionDetail()`](https://github.com/tommy0103/obelisk/blob/87ac8cbcfd8bf73eb491719683667a1b39c24ef6/packages/core/src/session-detail.ts#L408-L470) includes only `agent_id=null` messages in a first-class session's detail. I reproduced this with the real fixture: the two `subagent_child` sessions emit 6 and 10 messages, but both assemble to zero detail messages. Please cover opening a child session through the direct and persisted detail seams.

3. **[P1] Parent deletion/restoration does not update unchanged children.** [Discovery](https://github.com/tommy0103/obelisk/blob/a3e916aea7843796d21a74ea15cb471bc367f9c8/packages/core/src/providers/zcode.ts#L338-L377) skips a child when its own fingerprint is unchanged, before evaluating the parent's liveness for `isSubagent`. Deleting the parent emits only its tombstone; [persistence then removes](https://github.com/tommy0103/obelisk/blob/87ac8cbcfd8bf73eb491719683667a1b39c24ef6/packages/core/src/persist.ts#L210-L219) the parent-linked subagent row. Restoring the parent cannot recreate that link until the child itself changes. Please make parent-liveness changes trigger a child replay and test both transitions.

**Other review points**

4. **[P2] Textless rows are classified as text.** [This projection](https://github.com/tommy0103/obelisk/blob/a3e916aea7843796d21a74ea15cb471bc367f9c8/packages/core/src/providers/zcode.ts#L743-L764) writes `content_type: 'text'` even when `text` is null. The real fixture produces six visible rows of this shape, including tool-only responses. `CONTRIBUTING.md` requires `unknown` for textless records; `tool_use` is appropriate where the structured tool call is the content.

5. **[P2] The compact summary is stored but unavailable through standard retrieval.** The fixture's `compact_summary` has a hidden/visible/hidden visibility triad, so [this summary record](https://github.com/tommy0103/obelisk/blob/a3e916aea7843796d21a74ea15cb471bc367f9c8/packages/core/src/providers/zcode.ts#L597-L609) is `hidden`. `summaries({ includeInactive: true })` and session detail both exclude it. If it is intentionally hidden, please narrow the PR's claim about compact summaries; otherwise adjust the projection and exercise a public retrieval path.

6. **[P2] The source binding differs from ADR-0005.** [The adapter imports `node:sqlite` directly](https://github.com/tommy0103/obelisk/blob/a3e916aea7843796d21a74ea15cb471bc367f9c8/packages/core/src/providers/zcode.ts#L44-L48), whereas [ADR-0005](https://github.com/tommy0103/obelisk/blob/87ac8cbcfd8bf73eb491719683667a1b39c24ef6/docs/adr/0005-app-electron-vite-ts-esm.md) keeps the Core graph free of `node:sqlite` for the App and injects `better-sqlite3`. The installed Electron 43 runtime does expose `node:sqlite`, so I have not observed a startup failure; please either retain the documented binding seam or make this exception an explicit architecture decision with App runtime coverage.

